Issue #3, opened 2022-02-22
Docs: how-to guide for Qiskit-Braket plugin


IceKhan13:
### What is the expected behavior?
Create docs and example notebooks on how-to use Qiskit-Braket plugin.

References:
- https://github.com/aws/amazon-braket-examples/tree/main/examples/pennylane

IceKhan13:
Closing as we implement it in #11 #21 #24 and release in https://github.com/qiskit-community/qiskit-braket-provider/releases/tag/v0.0.1

PR #24, opened 2022-04-26
Refactor: rework of job result + adding measure instruction to target

IceKhan13:
### Summary
Refactor: rework of job result + adding measure instruction to target

Also tutorial notebook added

### Details and comments
Part of #3


IceKhan13:
@Morcu check this out too, please

PR #60, opened 2022-08-26
Add ECR gate to mappers

dexter2206:
### Summary

This adds ECR gate to mappers in the `adapters.py` module. Without this addition, running circuits containing the ECR gate through any of the AWS backends is impossible.

### Details and comments

No unit tests were added since there are already tests checking all the mappers. There are also no specific tests for other gates, so I wanted to stay consistent.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -29,6 +29,7 @@ from qiskit.circuit.library import (
     CSwapGate,
     CYGate,
     CZGate,
+    ECRGate,
     IGate,
     RXGate,
     RYGate,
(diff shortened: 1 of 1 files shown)


IceKhan13 (approved):
Sweet! Thank you for this addition @dexter2206 !

IceKhan13 (changes requested):
I think linter is not happy about missing commas at ends of lists :) 

Easy fix will be to run black `black qiskit_braket_provider`

dexter2206:
> I think linter is not happy about missing commas at ends of lists :)
> 
> Easy fix will be to run black `black qiskit_braket_provider`

Ah, sorry, I forgot to run black. Fixed in 9fb5c2a :)

IceKhan13 (approved):
Sweet! Thank you for contribution 🚀

Issue #45, opened 2022-06-15
Rename AWSBraketJob to AmazonBraketTask


JordanAWS:
https://github.com/qiskit-community/qiskit-braket-provider/blob/1f64aaeceb241df3eb10e696f1e5034e73c1d712/qiskit_braket_provider/providers/braket_job.py#L63

**Make sure that `AWSBraketJob` remains an alias for `AmazonBraketTask`** to avoid breaking existing code.

JordanAWS:
Will also need to edit all other references to `AWSBraketJob` in the rest of this repo. I'd recommend searching "AWSBraketJob" in your local IDE to find all references, since the search feature on GitHub often misses some.

IceKhan13:
@dexter2206 I like second option more 😄  if you do not mind doing this, it would be great!

I will assign it to you for now then. Thank you ❤️

devilkiller-ag:
Hi, Can I work on this issue?

devilkiller-ag:
Do I need to re run notebooks in doc to change the AWSBraketJob Name?

IceKhan13:
@devilkiller-ag yes, that would be great :) 
Do not forget about deprecation too

prakharb10:
Yeah, I encountered the same error. There are additional steps for running the code outlined here: https://aws.amazon.com/blogs/quantum-computing/introducing-the-qiskit-provider-for-amazon-braket/

Hence, the need for an AWS account

devilkiller-ag:
@IceKhan13 I have added Deprecation Warnings.

devilkiller-ag:
@IceKhan13 I tried making AWS Account, It's not accepting Indian Back Credit Card. It's particularly asking for MasterCard, Visa or AmEx. Is their any other way to rectify docs?

christianbmadsen:
> Is their any other way to rectify docs?

@devilkiller-ag we are assessing options for you to get unblocked. Will circle back with suggested path. Thanks!

christianbmadsen:
@devilkiller-ag we have identified next steps to get you unblocked. Could you send an email to unitary-hack-braket-support@amazon.com and we will follow-up with you? Thanks!

devilkiller-ag:
@christianbmadsen sure I am sending it

devilkiller-ag:
[...]

Issue #58, opened 2022-08-19
Supporting verbatim circuit execution


dexter2206:
### What is the expected enhancement?

I noticed that it is not currently possible to define a circuit in Qiskit and then run it on AWS device in verbatim mode and/or with disabled qubit rewiring. This feature seems to be very important, especially in the field of benchmarking quantum devices, when one needs to have precise control of what circuits are being executed. Is there any chance this could get implemented?

One possibility would be to introduce a `verbatim` and `disable_qubit_rewiring` options in `AWSBraketBackend.run` method, and then act adequately by wrapping the circuit in verbatim instruction before execution or adding `disable_qubit_rewiring=True`.

IceKhan13:
Hey @dexter2206 ! Thanks for the submission!

I think qubits rewiring is happening on transpilation step. So, if you do not transpile circuit and send it directly, then it should be executed as given. 

Maybe I'm wrong and verbatim is happening on QASM level?

Can you spill a little more light on it @dexter2206?

dexter2206:
Thanks for your response @IceKhan13 ! I think that unless `disable_qubit_rewiring=True` is passed, the rewiring is done by Braket automatically. Similarly, the circuit being sent is compiled for a given device unless it is explicitly wrapped in a verbatim box. 

Verification should be simple: I can send a circuit with incompatible gateset and/or incompatible qubit connectivity to a physical device and see if it is executed. If yes, the rewiring/compilation is happening automatically and we would need some explicit actions to prevent it. I will verify this once Lucy or Aspen M-2 are available in my region.

dexter2206:
Consider the following piece of code, which runs CNOT(0, 1) on Lucy device in verbatim mode.
```python
from braket.circuits import Circuit
from braket.aws import AwsDevice

subcircuit = Circuit().cnot(0, 1)
circuit = Circuit().add_verbatim_box(subcircuit)

device = AwsDevice("arn:aws:braket:eu-west-2::device/qpu/oqc/Lucy")
device.run(circuit, shots=10)
```
This code correctly raises an error, because Lucy device doesn't support CNOT gate natively. However, this cannot be reproduced using `qiskit_braket_provider`, because there is no way of adding verbatim box. By default, circuits without verbatim boxes are  compiled by Braket, and hence for instance the following code does not raise an error

```python
from qiskit import QuantumCircuit
from qiskit_braket_provider import AWSBraketProvider

[...]

PR #61, opened 2022-08-30
Add support for verbatim circuits and disable_qubit_rewiring

dexter2206:
### Summary
This PR implements support for verbatim boxes (only for whole circuits) and disabling qubit rewiring. The solution follows exactly the discussion in #58 

### Details and comments
Note that tests for the functionality have a `skip` mark because they necessarily have to access external resources.


IceKhan13 (commented):
@dexter2206 amazing work! All good on code side!

I only have comment on notebook. Last cell is not executed, so, we do not see result. And maybe we should catch error. By that we will show that error is expected. 

What do you think?

IceKhan13:
I think we merge this and fix https://github.com/qiskit-community/qiskit-braket-provider/issues/45 and we will cut new release.

dexter2206:
> @dexter2206 amazing work! All good on code side!
> 
> I only have comment on notebook. Last cell is not executed, so, we do not see result. And maybe we should catch error. By that we will show that error is expected.
> 
> What do you think?

Thanks! I will change the notebook as suggested tomorrow when Aspen is available.

IceKhan13:
@dexter2206 hi there! any updates here? :)

IceKhan13 (approved):

PR #66, opened 2022-11-22
Update retry to fix #65

### LINKED ISSUE #65: os.environ throws TypeError
The os.environ["QISKIT_BRAKET_PROVIDER_MAX_DELAY"] must be set as a string, meaning that when it is set and the _get_result_from_aws_tasks is called, then the retry decorator throws a TypeError since it cannot compare an int and a str.

TypeError is raised when os.environ["QISKIT_BRAKET_PROVIDER_MAX_DELAY"] is set.

Place an int() in the retry decorator to eliminate the problem.

stigerasmussen:
### Summary
Added int() around os.environ.
Test and documentation should not need update

### DIFF
--- a/qiskit_braket_provider/providers/braket_job.py
+++ b/qiskit_braket_provider/providers/braket_job.py
@@ -20,8 +20,8 @@ def retry_if_result_none(result):
 
 @retry(
     retry_on_result=retry_if_result_none,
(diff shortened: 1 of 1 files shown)


IceKhan13 (approved):
Neat catch! Thank you @stigerasmussen !

PR #68, opened 2022-11-22
Fix memory format

jcjaskula-aws:
### Summary

The Braket `result().measurement` field returns shot results using np.array(). Qiskit expect memory to be a list of bitstring. This MR modifies how the results are passed to ExperimentResultData.

This PR also fixes test_braket_job unittest.

### Details and comments
Closes #53 
Closes #59 (as memory=True is not needed)

Closes #67

PR #70, opened 2022-11-23
Fix result shots

jcjaskula-aws:
### Summary
This PR updates the way we construct ExperimentResult by filling the `shots` field by the value extracted from task metadata. 

### Details and comments
This changes the private API: `_get_result_from_aws_tasks` does not require the `shots` arguments anymore. 

It also reverts a change from #68 (where measurements was wrongly changed to list(array)).


IceKhan13 (approved):
LGTM!

Issue #47, opened 2022-06-15
add get_statevector support for BraketLocalBackend


JordanAWS:
Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/qiskit-community/qiskit-braket-provider?pullRequest=48) before we can accept your contribution.

Currently `BraketLocalBackend().run(circuit).result().get_statevector()` will fail. The state vector result type is available for the Braket Local Simulator: https://github.com/aws/amazon-braket-sdk-python/blob/main/examples/bell_result_types.py#L40

manulpatel:
Hello @JordanAWS! I new to qiskit-bracket and would like to work on this issue. Could you please assign it to me and provide if any further hints?

JordanAWS:
Thanks @manulpatel! Just assigned! Let me know if you have any specific questions or if the prompt is unclear :)

manulpatel:
Thanks @JordanAWS for assigning me! But it seems @IceKhan13 has already done the work. You may consider him.

JordanAWS:
Ah, good point. We will review his PR.

dakk:
> Ah, good point. We will review his PR. 
> 

Thank you Jordan

dakk:
Since the PR has been merged, should be the issue closed @JordanAWS ?

PR #95, opened 2023-05-27
Add get_statevector support

dakk:
### Summary

- Add get_statevector capability for shots=0 simulations on BraketLocalBackend (#47)

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -107,6 +107,8 @@ class BraketLocalBackend(BraketBackend):
         )
         circuits: List[Circuit] = list(convert_qiskit_to_braket_circuits(convert_input))
         shots = options["shots"] if "shots" in options else 1024
+        if shots == 0:
+            circuits = map(lambda x: x.state_vector(), circuits)
         tasks = []
         try:
             for circuit in circuits:
--- a/qiskit_braket_provider/providers/braket_job.py
+++ b/qiskit_braket_provider/providers/braket_job.py
@@ -41,16 +41,24 @@ def _get_result_from_aws_tasks(
     for task in tasks:
         if task.state() in AwsQuantumTask.RESULTS_READY_STATES:
             result: GateModelQuantumTaskResult = task.result()
-            counts = {
-                k[::-1]: v for k, v in dict(result.measurement_counts).items()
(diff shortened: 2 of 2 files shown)


kshitijc (commented):
Thank you for this change @dakk! Looks great! 🙌  Could you please:
* Fix the failing lint check
* Add an example [here](https://github.com/qiskit-community/qiskit-braket-provider/tree/main/docs/how_tos) along with documentation detailing this use case

dakk:
> Thank you for this change @dakk! Looks great! raised_hands Could you please:
> 
>     * Fix the failing lint check
> 
>     * Add an example [here](https://github.com/qiskit-community/qiskit-braket-provider/tree/main/docs/how_tos) along with documentation detailing this use case

Lint fixed and add a simple notebook for local simulations and statevector calculation. Let me know if it is ok

IceKhan13:
@dakk there is still one linter error. Check [details on CI](https://github.com/qiskit-community/qiskit-braket-provider/actions/runs/5132451858/jobs/9239138750?pr=95).

Basically in braket backend class `circuits` which has type `List[Circuit]` and you trying to assign type map to it :) 
Make sure types are matched and we are good to do :)

dakk:
> @dakk there is still one linter error. Check [details on CI](https://github.com/qiskit-community/qiskit-braket-provider/actions/runs/5132451858/jobs/9239138750?pr=95).
> 
> Basically in braket backend class `circuits` which has type `List[Circuit]` and you trying to assign type map to it :) Make sure types are matched and we are good to do :)

Sorry for that, now should be fixed; I surrounded map with list, so List as type is correct. I've some troubles with running mypy on my computer

docs/how_tos/5_how_to_run_circuits_on_Braket_local_backend.ipynb
```diff
@@ -0,0 +1,128 @@
+{
```
kshitijc, line 1: Would it make sense to merge with [this](https://github.com/dakk/qiskit-braket-provider/blob/get_statevector/docs/how_tos/2_how_to_retrieve_results_from_backend.ipynb) how-to, since the primary motivation is to show how to retrieve a specific result (state vector) here?

kshitijc (approved):
Nice work @dakk! 🚀

PR #97, opened 2023-05-29
Updates for Qiskit to Braket circuit conversion

### LINKED ISSUE #37: Updates for Qiskit to Braket circuit conversion
Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/qiskit-community/qiskit-braket-provider?pullRequest=48) before we can accept your contribution.

Currently the [convert_qiskit_to_braket_circuit](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/qiskit_braket_provider/providers/adapter.py#L328-L334) function:
* *Uses Probability result type for the Qiskit measure instruction*: It might be better to use the Sample result type if you wish to retrieve the measurement values (the raw 0/1 values). Probability result type might in the future choose to omit the raw measurements. 
* *Doesn't support the barrier instruction*: Look at [verbatim support](https://docs.aws.amazon.com/braket/latest/developerguide/braket-constructing-circuit.html#verbatim-compilation) and see if this aligns with barrier.

urihan:
### Summary
resolves #37 by changing Probability result type to Sample result type. 

### Details and comments
I have added an additional class, Observable, to the adapter.py file. Also updated the Sampler result type itself by adding an observable z as a parameter.


kshitijc (commented):
Thank you for this contribution @urihan! 🚀  Could you please make changes to:
* Fix the failing checks
* Add unit tests for the source code changes

qiskit_braket_provider/providers/adapter.py
```diff
@@ -347,11 +347,12 @@ def convert_qiskit_to_braket_circuit(circuit: QuantumCircuit) -> Circuit:
             # TODO: change Probability result type for Sample for proper functioning # pylint:disable=fixme
             # Getting the index from the bit mapping
             quantum_circuit.add_result_type(
-                result_types.Probability(
+                result_types.Sample(
+                    observable=Observable,
```
dakk, line 351: Class `Observable` represents a quantum observable, but which one? You have to pass the computational basis observable

urihan: Hi @dakk, Thank you for your comment. In my first commit, I used `observable=Observable.Z()` for the Probability; however, it was not passing the `tox -elint` test. I was changing codes to pass the `tox -elint`, which is why I got rid of the `Z()` part.

dakk: What was the error of the linter?

urihan: @dakk qiskit_braket_provider/providers/adapter.py:351:31: E1101: Class 'Observable' has no 'Z' member (no-member)

urihan: I wonder if we have to use `StandardObservable` class instead of just `Observable` class

dakk: No, it's correct to use `Observable.Z()`; maybe your braket version is outdated. This is my env, and Observable.Z is working:

image

urihan: Right, seems like I have the same version. For that specific error, I added the comment `# pylint:disable=fixme` next to the code, and I think it can fix it. 

But, now I am getting after running `tox -elint` 

```
lint run-test: commands[3] | mypy --install-types --non-interactive .
error: --install-types failed (no mypy cache directory)
```
I tried to follow this instruction on [mypy-install-types-mre](https://github.com/tk-woven/mypy-install-types-mre), but only produces the same error. I might have to spend more time with it.

dakk: Maybe there is a problem with the linter that can't detect dynamically added members; observables are dynamically added to Observable class (from braket side).

There is a workaround:

```python
from braket.circuits import observables
observables.Z()
```

If the problem is the linter, this should work.

urihan:
Hi @kshitijc, I have finally fixed failing checks and also added a unit test! It would be great if you can check my submission. Thank you!

kshitijc (approved):
Great job @urihan! 👍

qiskit_braket_provider/providers/adapter.py
kshitijc: Thanks a lot for your help here @dakk!

dakk: > I wrote a comment here before, but I might have accidentally erased it. Thank you for your help @dakk. I found out I was still getting the same error even after updating with observables.Z(). I was able to get it to work after re-cloning my repository. I decided to just stick with the solution you provided though! I appreciate your help! :)

Oh, that's weird! It was a pleasure!

Issue #90, opened 2023-05-23
Gate decomposition


speller26:
Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/qiskit-community/qiskit-braket-provider?pullRequest=48) before we can accept your contribution.

Right now, the Amazon Braket provider translates gates from Qiskit to Braket with a [hardcoded dicts](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/qiskit_braket_provider/providers/adapter.py#L58-L159). This limits the Qiskit operations that can be used in Braket. For example, the circuit

```python
circuit = QuantumCircuit(1)
circuit.prepare_state([1/np.sqrt(2), -1/np.sqrt(2)], 0)
```

cannot run because `StatePreparation` is not explicitly in the dict. Likewise,

```python
operator = (Z ^ Z) - 0.1 * (X ^ I)
evo = PauliEvolutionGate(operator, time=0.2)
circuit = SuzukiTrotter().synthesize(evo)
```

cannot run because of the absence of exponential operators. These circuits _can_ run on Braket provided they are sufficiently decomposed.

Another example of an unsupported gate is the `U` gate;

```python
circuit = QuantumCircuit(1)
circuit.u(np.pi/2, np.pi/3, np.pi/4, 0)
```

cannot run, regardless of how many times `decompose()` is called, because `U` is the final step of the decomposition; for gates such as these, the decomposition needs to be handled directly.

### What is the expected behavior?

Gats should be automatically decomposed to be compatible with Amazon Braket devices, and gates that cannot be made compatible with (built-in) decomposition should be converted in the provider. 

### Acceptance criteria
A circuit with any [standard gate](https://qiskit.org/documentation/apidoc/circuit_library.html#standard-gates) (aside from global phase) should run successfully, as well as each of the circuits described above; these should all be included as test cases.

abrassel:
I am also interested in working on this issue.

Gmontes01:
Hello, I would also like to be assigned to this project @speller26

s-aldaihan:
I am also interested in this issue @speller26

JordanAWS:
Since multiple people (@s-aldaihan , @Gmontes01 , @AbdullahKazi500 )  have volunteered for this issue, I will assign it to all of you. 

Here are the rules: you can either 
A) Work together on this issue and split the bounty (very reasonable idea for a project of this size!)
[...]

PR #106, opened 2023-06-09
Gate Decomposition(#90) Added and Tested 

Gmontes01:
I implemented the asked changes in issue Gate Decomposition(#90). I added a step in the qiskit to braket converter which decomposes the qiskit circuit to a form which can be translated to braket using the built in dictionary. 

The changes test and handle the prepare_state() method for preparing qubit states
```
    circuit = QuantumCircuit(1)
    circuit.prepare_state([1/np.sqrt(2), -1/np.sqrt(2)], 0) 
```(note that a given qubit needs to start in the 0 state for the prepare_state() method to work properly)
and tests and handles .u() gates added to qiskit circuits like in the code below
```
 circuit = QuantumCircuit(1)
circuit.u(np.pi/2, np.pi/3, np.pi/4, 0)
```
and the code can decompose exponential gates like the ones listed below
```
operator = (Z ^ Z) - 0.1 * (X ^ I)
evo = PauliEvolutionGate(operator, time=0.2)
circuit = SuzukiTrotter().synthesize(evo)
```
All of these codes have their own tests added to the test_adapter.py file.

Also note the test added to test_adapter.py which checks that all standard gates are decomposed into a circuit which can be 
translated by the hardcoded dictionary.

### Details and comments
Also note that the decomposition maintains the overall phase of the circuit by correcting the phase if the decomposition ends up creating a phase shift. 
I also changed the qiskit to braket gate map to map 'u' and 'u3' to 
```
gates.Rz(lam),
gates.Ry(theta),
gates.Rz(phi),
gates.PhaseShift((lam + phi) * (0.5)),
gates.X(),
gates.PhaseShift((lam + phi) * (0.5)),
gates.X(),
``` 
instead of 
```
gates.Rz(lam),
gates.Rx(pi / 2),
gates.Rz(theta),
gates.Rx(-pi / 2),
gates.Rz(phi)
``` 
in order to prevent the global phase shift which was taking place in the later code which interfering with my testing of the adapter's handling of the 'u' gate. Also note that this change required me to change the parametric circuit conversion test and its expected value since the test checked the output of translations of the 'u' gate with an expected circuit.

### CODE REVIEW

File: qiskit_braket_provider/providers/adapter.py [hunk 1]
```diff
@@ -350,6 +356,36 @@ def aws_device_to_target(device: AwsDevice) -> Target:
...
+    translatable_gates = set(qiskit_gate_names_to_braket_gates.keys())
+    translatable_gates = translatable_gates.union({"measure", "barrier"})
+    # decompose quantum circuit
+    while not (
+        {gate.name for gate, _, _ in decomposed_circuit.data}.issubset(
+            translatable_gates
+        )
+    ):
+        old_circuit = decomposed_circuit
+        decomposed_circuit = old_circuit.decompose()
```
kshitijc, line 377: Is this always guaranteed to terminate?

Gmontes01: Reasonably this should terminate for circuits that are composed of standard gates, and measure/barrier instructions since the adapter passes the standard gates and paramterized circuit test. 

This may pose a potential bug though if my assumption is not correct and I can mitigate it by adding a condition where if decomposing the gate does nothing then the loop breaks.

Issue #88, opened 2023-05-23
Create Qiskit `QuantumCircuit` from Braket `Circuit` objects


kshitijc:
### What is the expected behavior?

Currently, the Qiskit Braket Provider allows you to convert Qiskit `QuantumCircuit` objects to Braket `Circuit` objects. It would be useful to be able to convert Braket `Circuit` objects to Qiskit `QuantumCircuit` objects, allowing users to make use of any of the utilities available in Qiskit to visualize or transform the `QuantumCircuit` before optionally translating it back to a Braket `Circuit` if needed.

The method:
```
def from_braket_circuit(circuit: Circuit) -> QuantumCircuit:
   ...
```

could potentially leverage the existing `from_qasm_str` method available in the QuantumCircuit class or operate directly on the Braket Circuit objects, and could also be potentially exposed on the Quantum circuit class to have a similar usage pattern as the `from_qasm_str` method:
```
QuantumCircuit.from_braket_circuit = from_braket_circuit
```

#### Acceptance criteria
* The implementation should be able to handle the Braket circuit containing the following appropriately:
  * Supported [gates](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/gates.py)
  * Supported [result types](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/result_types.py) and [observables](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/observables.py)
  * Supported [noise instructions](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/noises.py)
  * [Verbatim blocks](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/compiler_directives.py)
* The implementation should be accompanied by unit tests with full coverage.
* The implementation method should be accompanied by an example in [this folder](https://github.com/qiskit-community/qiskit-braket-provider/tree/main/docs/how_tos).

ogunsegun:
https://github.com/ogunsegun/Rebulid-of-Create-Qiskit-QuantumCircuit-from-Braket-Circuit-objects.git

JordanAWS:
@ogunsegun Can you submit a pull request on the main repo using a feature branch (rather than your fork)?

ogunsegun:
plz how am my going to do that @JordanAWS 
because the is in my github

SimoneGasperini:
[...]

PR #112, opened 2023-06-14
Creating `Qiskit QuantumCircuit` from `Braket Circuit` objects (related to issue #88)

denvitko:
### Summary

Creating `Qiskit QuantumCircuit` from `Braket Circuit` objects. See issue #88.

### Details and comments

Under construction.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -426,8 +426,14 @@ def from_braket_circuit(circuit: Circuit) -> QuantumCircuit:
     num_qubits = circuit.qubit_count
     quantum_circuit = QuantumCircuit(num_qubits)
     
-    for instruction in circuit.instructions:
-        print(instruction)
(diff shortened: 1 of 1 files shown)


laurencap:
Hi @denvitko! Thank you for opening this PR and contributing to qiskit-braket-provider. We appreciate it! While this is still a work in progress, would you mind converting the PR to a draft PR? You can do that by clicking "Convert to draft" under "Reviewers" on the right side of the page.

Thank you for commenting on the issue as well! Let's continue to use that for discussion and questions related to this work.

denvitko:
The function `from_braket_circuit()` is now able to handle most of Braket gates without parameter (including multi-qubit control gates). Gates that need additional implementation are: `cphaseshift00`, `cphaseshift01`, `cphaseshift10`, `pswap`.

I would like to continue with implementing parametrized gates (so far it's possible to use parameters, but there must be different parameter for each gate), and power gates. Last but not least, it should be taken care of all unitary gates in general.

PR #123, opened 2023-10-06
feat: add queue information details for devices and jobs

virajvchaudhari:
### Summary
- Add queue depth information for the Aws Devices and queue position information for the submitted jobs. 
- Add tests and checked for lint issues.

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -7,6 +7,7 @@ from abc import ABC
 from typing import Iterable, Union, List
 
 from braket.aws import AwsDevice, AwsQuantumTaskBatch, AwsQuantumTask
+from braket.aws.queue_information import QueueDepthInfo
 from braket.circuits import Circuit
 from braket.devices import LocalSimulator
 from braket.tasks.local_quantum_task import LocalQuantumTask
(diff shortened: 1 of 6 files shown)


qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -214,6 +215,39 @@ def qubit_properties(
         # TODO: fetch information from device.properties.provider  # pylint: disable=fixme
         raise NotImplementedError
 
+    def queue_depth(self) -> QueueDepthInfo:
```
kshitijc, line 218: The Qiskit backend interface expects this information in the [BackendStatus](https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/models/backendstatus.py#L27) object. Could we fit this in there?

qiskit_braket_provider/providers/braket_job.py
```diff
@@ -119,6 +120,42 @@ def shots(self) -> int:
     def submit(self):
         return
 
+    def queue_position(self) -> QuantumTaskQueueInfo:
```
kshitijc, line 123: queue_position and queue_info are two separate methods for IBM Jobs [[Ref](https://qiskit.org/documentation/stable/0.19/_modules/qiskit/providers/ibmq/job/ibmqjob.html#IBMQJob.queue_position)] Is it possible for us to align to make it easier for qiskit customers to easily switch between IBM and Braket?

qiskit_braket_provider/providers/braket_backend.py
virajvchaudhari: Qiskit providers uses [BackendV2](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/qiskit_braket_provider/providers/braket_backend.py#L32), while the qiskit backend uses [BackendV1](https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/backend.py#L40), so this change would be incompatible.

kshitijc (approved):

PR #127, opened 2023-10-21
doc: update hybrid job tutorial to use base container instead of byoc

guomanmin:


christianbmadsen (commented):
Thank you for this PR and step to simplify and improve this example!

Title and introduction:

We should assume a person seing this is familiar with Qiskit and unfamiliar with Braket. The introduction needs to take that context into account. A good title may be something like "Running variational workloads with qiskit on a mix of AWS ec2 instances and available quantum devices". The intro similarly should be expanded. Maybe something like

"Amazon Braket Hybrid Jobs offers a way for you to run hybrid quantum-classical algorithms that require both classical resources and quantum processing units (QPUs). Hybrid Jobs is designed to spin up the requested classical compute, run your algorithm, and release the instances after completion so you only pay for what you use. This workflow is ideal for long-running iterative algorithms involving both classical and quantum resources. In this notebook we show you how to run such a job via the qiskit-braket provider."

Preparatory steps:

We need to explain what a job script is for. Something around the fact that the ec2 instance will be runnning classical code which is packaged as a "job script". You also need to explain what is happening in the job script, not just have the code ("we are seting up a VQE script similar to and this script will run the optimization on AWS ec2 and run the quantum circuits on the AWS on-demand simulator SV1 ")

Running your job:

Change "The Braket-managed base container have qiskit-braket-provider preinstalled and ready to use for running your variational algorithms" to "Amazon Braket provides a container for Hyrbid Jobs that has the qiskit-braket container and associated dependencies installed."

Add a sentence of what the results show/mean. A plot would be even better.

docs/tutorials/2_tutorial_hybrid_jobs.ipynb
```diff
@@ -0,0 +1,263 @@
...
+    "In summary, this script sets up the hybrid job script that executes a VQE computation using a the Amazon Braket SV1 default simulator as the quantum backend and then saves the results of the computation to the EC2 instance the hybrid job is running on."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6010160f",
+   "metadata": {},
+   "source": [
+    "## Running your job\n",
+    "Amazon Braket provides a container for Hyrbid Jobs that has the qiskit-braket container and associated dependencies installed. To use the base container, call `retrieve_image(Framework.BASE, AwsSession().region)` as demonstrated below:"
```
zmohammad01, line 133: Hybrid Jobs* (spelling)

docs/tutorials/2_tutorial_hybrid_jobs.ipynb
```diff
@@ -0,0 +1,263 @@
...
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "b9184273",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\"\"\"Example of Hybrid Job payload with VQE.\"\"\"\r\n",
```
zmohammad01, line 63: Shall we explain somewhere at the top what exactly a VQE is, why it needs hybrid jobs. Then at the bottom we can explain what these results mean.

docs/tutorials/2_tutorial_hybrid_jobs.ipynb
guomanmin: Addressed

PR #133, opened 2023-12-26
Use the U gate to translate Qiskit U-like gates

jcjaskula-aws:
### Summary
Replace `U`s decomposition by the application of a the Braket U gate.

### Details and comments
Since U has been added to the Braket SDK, we simplify the implementation here.

Bump qiskit-terra version to solve an occurence of https://github.com/Qiskit/qiskit/issues/7326


ajberdy (approved):
Approving to merge post 3.8 deprecation

qiskit_braket_provider/providers/adapter.py
```diff
@@ -76,22 +76,10 @@
...
-        braket_gates.PhaseShift(phi),
-    ],
-    "u": lambda theta, phi, lam: [
-        braket_gates.PhaseShift(lam),
-        braket_gates.Ry(theta),
-        braket_gates.PhaseShift(phi),
-    ],
+    "u1": lambda lam: [gates.U(0, 0, lam)],
+    "u2": lambda phi, lam: [gates.U(pi / 2, phi, lam)],
+    "u3": lambda theta, phi, lam: [gates.U(theta, phi, lam)],
```
speller26, line 81: Since the backend now [looks at the device's supported gates](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/qiskit_braket_provider/providers/braket_backend.py#L123-L127), `u1`, `u2` and `u3` will never be included in the basis gates for `transpile`; you'll need to add a condition after the linked lines to add these three gates if `U` is present.

speller26:
I think this is actually still worth finishing without waiting for the service; the service doesn't support the `U` gate, but the same goes for control modifiers, which _are_ supported in the provider.

speller26 (approved):
> I'm not sure the service handles more gracefully `U` gates than control modifiers. We could try and fix it if it is a real issue.

The way I see it, since the service supports neither control modifiers nor $U$ gates, but we've introduced the former anyway due to its support in the local simulator, we might as well do the same for the latter.

This can also help remind users of the full capabilities of the local simulator.

PR #137, opened 2024-01-11
raise type error for bad adapter circuit conversion input type

ryanhill1:
### Summary

More descriptive error message for bad conversion function input.

Before:

```python
from qiskit_braket_provider.providers.adapter import convert_qiskit_to_braket_circuit

convert_qiskit_to_braket_circuit("abc")
# AttributeError: 'Circuit' object has no attribute 'data'
```
```python
from unittest.mock import Mock

bad_input = Mock()
bad_input.data = []

convert_qiskit_to_braket_circuit(fake)
# TypeError: '>' not supported between instances of 'Mock' and 'float'
```
etc.

Now:

```python
convert_qiskit_to_braket_circuit(input)
# TypeError: 'Expected a qiskit.QuantumCircuit, got <type(input)> instead'
```

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -416,6 +416,11 @@ def convert_qiskit_to_braket_circuit(circuit: QuantumCircuit) -> Circuit:
     Returns:
         Circuit: Braket circuit
     """
+    if not isinstance(circuit, QuantumCircuit):
+        raise TypeError(
+            f"Expected a qiskit.QuantumCircuit, got {type(circuit)} instead"
+        )
+
     quantum_circuit = Circuit()
     if not (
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -416,6 +416,11 @@ def convert_qiskit_to_braket_circuit(circuit: QuantumCircuit) -> Circuit:
     Returns:
         Circuit: Braket circuit
     """
+    if not isinstance(circuit, QuantumCircuit):
+        raise TypeError(
+            f"Expected a qiskit.QuantumCircuit, got {type(circuit)} instead"
```
math411, line 421: Minor nit: this does line up with the type hint. If others feel this is more clear than just `QuantumCircuit`, I am fine with keeping it.

ryanhill1: Either works for me, happy to change it lmk

PR #139, opened 2024-01-26
Enable use of qiskit ParameterExpressions for circuit conversion: qiskit to braket

lukasvoss:
### Summary

- Users that work with a qiskit circuit that was parameterized using a qiskit ParameterExpression can now convert their qiskit circuit to a braket circuit using the ``convert_qiskit_to_braket_circuit`` function.

### Details

- A qiskit ParameterExpression could be for example a qiskit Parameter object with a constant (i.e., float) offset. This could can be relevant is one uses a constant baseline for a rotation value on which a variable Parameter then gets added.

This could look something like this for a parametrized CNOT gate in qiskit:

```
q_reg = QuantumRegister(2)
my_qc = QuantumCircuit(q_reg, name="custom_cx")

baseline = np.pi * np.array([0.1, 0.3, 0.7, 0.3, -0.4, 0.4, -0.3])
params = ParameterVector(name='a', length=7)

my_qc.u(
    baseline[0] + params[0], # This would constitute a qiskit ParameterExpression
    baseline[1] + params[1],
    baseline[2] + params[2],
    q_reg[0],
)
my_qc.u(
    baseline[3] + params[3],
    baseline[4] + params[4],
    baseline[5] + params[5],
    q_reg[1],
)
my_qc.rzx(baseline[6] + params[6], q_reg[0], q_reg[1])
```

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -3,7 +3,6 @@ from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 import warnings
 
 from braket.aws import AwsDevice
-from qiskit.circuit import ParameterExpression
 from braket.circuits import (
     Circuit,
     FreeParameter,
@@ -30,7 +29,7 @@ from numpy import pi
 
 from qiskit import QuantumCircuit, transpile
 from qiskit.circuit import Instruction as QiskitInstruction
-from qiskit.circuit import Measure, Parameter
+from qiskit.circuit import Measure, Parameter, ParameterExpression
 from qiskit.circuit.library import (
     CCXGate,
     CPhaseGate,
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -3,6 +3,7 @@ from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 import warnings
 
 from braket.aws import AwsDevice
+from qiskit.circuit import ParameterExpression
 from braket.circuits import (
     Circuit,
     FreeParameter,
@@ -451,6 +452,8 @@ def convert_qiskit_to_braket_circuit(circuit: QuantumCircuit) -> Circuit:
             for i, param in enumerate(params):
                 if isinstance(param, Parameter):
                     params[i] = FreeParameter(param.name)
+                elif isinstance(param, ParameterExpression): # Enables users to use ParameterVector elements
+                    params[i] = FreeParameter(name=list(param._names.keys())[0]) # Name of the parameter is stored as a dict key
 
             for gate in qiskit_gate_names_to_braket_gates[gate_name](*params):
                 instruction = Instruction(


laurencap:
Hi @lukasvoss, thank you for the contribution! This is a great feature. Please add some tests that cover the lines you've added and show that your changes work correctly.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -451,6 +451,8 @@ def convert_qiskit_to_braket_circuit(circuit: QuantumCircuit) -> Circuit:
             for i, param in enumerate(params):
                 if isinstance(param, Parameter):
                     params[i] = FreeParameter(param.name)
+                elif isinstance(param, ParameterExpression): # Enables users to use ParameterVector elements
+                    params[i] = FreeParameter(name=list(param._names.keys())[0]) # Name of the parameter is stored as a dict key
```
laurencap, line 455: Possible issue here: `list(param._names.keys())[0]` since this is returning the dictionary keys, converting to a list, and then selecting the first one. There's no guarantee that a dictionary will return its keys in the same order every time, so it's possible this will fail at some point! There is likely a safer way to get the information you need here. If you need help, let me know and I'll take a look.

Nit: No need to include comments on these lines.

laurencap (changes requested):

qiskit_braket_provider/providers/adapter.py
lukasvoss: I have implemented the changes regarding how to name the parameter expression and ensure that it is compatible with QASM 3 to successfully run the circuit with bounded parameters.
By default, qiskit names elements of a ``ParameterVector`` are named like this 'a[<index>]'. The '[]' parentheses are not supported by the QASM 3 syntax which leads to an error when trying to run a Braket Circuit with free parameters following the '[]'-naming. Therefore, I ensured that these will be removed before the ``ParamterExpression`` is passed to create a FreeParameterExpression out of it.

lukasvoss: I am not familiar with GitHub Actions and have not seen a way how I can run the tests on my PR myself. Can you please assist here and let me know how I can run these tests?

lukasvoss:
Dear Lauren, dear Jean-Christophe,
I have implement some changes that are solving the errors arising from the qiskit naming convention of ``ParameterVector`` elements by removing the squared-parenthesis from the parameter name.

lukasvoss:
@jcjaskula-aws Thanks for your contribution, looks good to me!

qiskit_braket_provider/providers/adapter.py
```diff
@@ -466,11 +477,32 @@ def to_braket(
...
     return params
 
 
+def _rename_param_vector_element(parameter):
+    param_name = str(parameter._symbol_expr)
+    return f"{param_name.replace('[', '_').replace(']', '')}"
+
+
+def _validate_name_conflicts(parameters):
+    renamed_parameters = {_rename_param_vector_element(param) for param in parameters}
```
speller26, line 498: `_rename_param_vector_element` is called twice for each parameter, first in the validation, and again in the translation; can the output be reused?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -466,11 +477,32 @@ def to_braket(
...
     for i, param in enumerate(params):
-        if isinstance(param, Parameter):
+        if isinstance(param, ParameterVectorElement):
+            cleaned_param_name = _rename_param_vector_element(param)
+            params[i] = FreeParameter(cleaned_param_name)
+        elif isinstance(param, Parameter):
             params[i] = FreeParameter(param.name)
+        elif isinstance(param, ParameterExpression):
+            cleaned_param_name = _rename_param_vector_element(param)
+            params[i] = FreeParameterExpression(sympify(cleaned_param_name))
```
speller26, line 487: Would it make sense to use
```suggestion
            params[i] = FreeParameterExpression(cleaned_param_name.sympify())
```
instead

qiskit_braket_provider/providers/adapter.py
```diff
@@ -466,11 +476,31 @@ def to_braket(
...
             params[i] = FreeParameter(param.name)
+        elif isinstance(param, ParameterExpression):
+            renamed_param_name = _rename_param_vector_element(param)
+            params[i] = FreeParameterExpression(renamed_param_name)
+
     return params
 
 
+def _rename_param_vector_element(parameter):
+    return str(parameter).replace("[", "_").replace("]", "")
```
speller26, line 492: This actually doesn't work if the expression contains division of or by free parameters, but `sympify()` does; maybe it's worth supporting division first and then adding a test to this PR.

speller26: On second thought, this extends to other operations like `exp` and trigonometric operations; it's fine to leave as is and let the Braket SDK handle exceptions.

speller26 (approved):

lukasvoss (commented):
Will now be handled with renaming parameter values, adjusting for QASM3 compatibility by removing parenthesis and controlling for duplicates in the parameter naming.

lukasvoss:
@all: Thanks for all your contributions!

PR #142, opened 2024-02-02
bugfix: Handle `meas_level` from backend.run

laurencap:
### Summary

Validate meas_level if it is supplied to run, and if it's valid, remove it. AwsQuantumTask does not accept `meas_level` argument.

Note:
meas_level=0 is raw results, an array of complex values (think sampled data collected from an ADC)
meas_level=1 is kerneled results, so a single complex value
meas_level=2 is classified results, 0 or 1

meas_level is also often supplied as an enum. For example: https://github.com/Qiskit-Extensions/qiskit-experiments/blob/f22bb7b67893a1b75f7d0f1fecdb0bd63284e244/qiskit_experiments/framework/base_experiment.py#L454

### DIFF
--- a/tests/providers/test_braket_backend.py
+++ b/tests/providers/test_braket_backend.py
@@ -156,7 +156,7 @@ class TestAWSBraketBackend(TestCase):
         circuit.h(0)
         circuit.measure(0, 0)
         with self.assertRaises(exception.QiskitBraketException):
-            backend.run(circuit, shots=10, meas_level=0)
+            backend.run(circuit, shots=10, meas_level=1)
 
     def test_vqe(self):
         """Tests VQE."""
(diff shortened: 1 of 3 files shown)


tests/providers/test_braket_backend.py
```diff
@@ -141,6 +141,23 @@ def test_local_backend_circuit_shots0(self):
...
+        backend.run(circuit, shots=10, meas_level=2)
+
+    def test_meas_level_1(self):
+        """Check that there's an exception for asking for raw measurement results."""
+        backend = BraketLocalBackend(name="default")
+        circuit = QuantumCircuit(1, 1)
+        circuit.h(0)
+        circuit.measure(0, 0)
+        with self.assertRaises(exception.QiskitBraketException):
+            backend.run(circuit, shots=10, meas_level=0)
```
yitchen-tim, line 159: nit: should it be meas_level=1, according to the function name?

laurencap: ty!

yitchen-tim (approved):

kshitijc:
Changes look good to me. Please fix the failing style checks.

PR #144, opened 2024-02-07
Add support for IonQ native gates

jcjaskula-aws:
### Summary
We add support for the GPi, GPi2, MS gates. We use a direct map between Braket and Qiskit implementation.

### DIFF
--- a/tests/providers/mocks.py
+++ b/tests/providers/mocks.py
@@ -118,12 +118,7 @@ MOCK_GATE_MODEL_SIMULATOR_CAPABILITIES_JSON = {
             "actionType": "braket.ir.jaqcd.program",
             "version": ["1"],
             "supportedOperations": ["H", "CNOT"],
-        },
-        "braket.ir.openqasm.program": {
-            "actionType": "braket.ir.openqasm.program",
-            "version": ["1"],
-            "supportedOperations": ["H", "CNOT"],
-        },
+        }
     },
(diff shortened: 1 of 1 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -119,6 +123,13 @@
     "ryy": lambda angle: [braket_gates.YY(angle)],
     "ecr": lambda: [braket_gates.ECR()],
     "iswap": lambda: [braket_gates.ISwap()],
+    # IonQ gates
+    "gpi": lambda angle: [braket_gates.GPi(angle)],
+    "gpi2": lambda angle: [braket_gates.GPi2(angle)],
+    "ms": lambda angle_1, angle_2, angle_3: [
+        braket_gates.MS(angle_1, angle_2, angle_3)
+    ],
+    "zz": lambda angle: [braket_gates.ZZ(angle)],
```
speller26, line 132: Shouldn't the angles all be multiplied by $2 \pi$, since [IonQ's gate definitions](https://github.com/qiskit-community/qiskit-ionq/blob/main/qiskit_ionq/ionq_gates.py) use turns while the Braket SDK's use [angles](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/main/src/braket/circuits/gates.py)?

speller26 (approved):

PR #147, opened 2024-02-09
Include all gates in simulator targets

speller26:
### Summary

* Simulator targets now include all gates, regardless of qubit count
* Major refactor to make code more readable, with functions being extracted as appropriate
* Adds the device qubit count to the target.

### Details and comments

Prior to this change, simulator targets and QPU targets alike were limited to up to 2q gates. Although no Braket devices are subject to these restrictions, this restriction still makes sense for QPUs because there is no analog of coupling map (and corresponding properties) for 3+ qubits. However, Braket simulators do not have a notion of coupling map, and gates can be applied to any qubit without impacting result quality. As such, this change removes this restriction for simulators.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -300,7 +300,7 @@ def _qpu_target(
     for operation in action_properties.supportedOperations:
         instruction = _GATE_NAME_TO_QISKIT_GATE.get(operation.lower(), None)
 
-        # TODO: Add 3+ qubit gates when target supports them  # pylint:disable=fixme
+        # TODO: Add 3+ qubit gates when Target supports them  # pylint:disable=fixme
         if instruction and instruction.num_qubits <= 2:
             if instruction.num_qubits == 1:
                 target.add_instruction(
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -1,5 +1,7 @@
 """Util function for provider."""
+
 from collections.abc import Callable, Iterable
+from math import pi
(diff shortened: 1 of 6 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -232,210 +224,160 @@ def _get_controlled_gateset(max_qubits: Optional[int] = None) -> set[str]:
...
 
-        for instruction in instructions:
-            simulator_instruction_props: Optional[
-                dict[
-                    Union[tuple[int], tuple[int, int]],
-                    Optional[InstructionProperties],
-                ]
-            ] = {}
-            # adding 1 qubit instructions
+        # TODO: Add 3+ qubit gates when target supports them  # pylint:disable=fixme
```
speller26, line 303:
```suggestion
        # TODO: Add 3+ qubit gates when Target supports them  # pylint:disable=fixme
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -232,210 +224,160 @@ def _get_controlled_gateset(max_qubits: Optional[int] = None) -> set[str]:
...
 
-        for instruction in instructions:
-            simulator_instruction_props: Optional[
-                dict[
-                    Union[tuple[int], tuple[int, int]],
-                    Optional[InstructionProperties],
-                ]
-            ] = {}
-            # adding 1 qubit instructions
+        # TODO: Add 3+ qubit gates when Target supports them  # pylint:disable=fixme
```
speller26, line 303:
```suggestion
        # TODO: Add 3+ qubit gates once Target supports them  # pylint:disable=fixme
```

PR #159, opened 2024-02-20
Fix some Parameter translation edge cases

jcjaskula-aws:
### Summary
The transpiler sometimes returns payloads that we don't translate well. This PR fixes a couple of issues I ran into.

### Details and comments
- Parametric gates with FreeParameter are transpiled with parametric global phase, which was throwing an error at translation.
- `Parameter`s can sometimes be composed of only single numerical values, which are not translated to a `FreeParameter`s.

### DIFF
--- a/tests/providers/test_adapter.py
+++ b/tests/providers/test_adapter.py
@@ -4,9 +4,10 @@ from unittest.mock import Mock, patch
 
 import numpy as np
 import pytest
-from braket.circuits import Circuit, FreeParameter, Gate, Instruction, observables
+from braket.circuits import Circuit, Gate, Instruction, observables
 from braket.circuits.angled_gate import AngledGate, TripleAngledGate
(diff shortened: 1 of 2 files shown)


tests/providers/test_adapter.py
```diff
@@ -240,17 +234,29 @@ def test_ionq_gates(self):
 
     def test_global_phase(self):
```
yitchen-tim, line 235: This test iterates two test cases. `@pytest.mark.parametrize` may be more appropriate here.

yitchen-tim:
> Parameter can sometimes be composed of only a single numerical values, which is not translated to a FreeParameter.
Could you elaborate on this point a bit? Do you mean the qiskit's parameter is a single parameter instead of a List?  If so, I'm not sure which part fixes this.

yitchen-tim (approved):

Issue #89, opened 2023-05-23
Support noise models


speller26:
In Qiskit, it is possible to add a [noise model](https://qiskit.org/documentation/tutorials/simulators/3_building_noise_models.html#Adding-errors-to-a-Noise-Model) to a simulator device so that the noise model is applied to any circuit [run on the device](https://qiskit.org/documentation/tutorials/simulators/3_building_noise_models.html#Executing-the-noisy-simulation). Analogously, in the Amazon Braket SDK, it is possible to build [noise models](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/noise_model/noise_model.py#L84) that can be applied to circuits to make them into noisy circuits.

### What is the expected behavior?

Amazon Braket DM1 and local density matrix backends should be able to take a Qiskit noise model that will be applied to all circuits run on the device.

### Acceptance criteria

* Adding a `noise_model` kwarg to the Braket provider's `get_backend` method that attaches a noise model to be applied to every circuit run
* Adding `noise_model` and `braket_noise_model` properties to the Braket backends that allow users to inspect the Qiskit and Braket noise models, respectively

```python
dm1 = provider.get_backend("DM1", noise_model=noise_bit_flip)

# Run noisy circuit
circuit = ...  # noiseless circuit
task = dm1.run(circuit, shots=100)  # circuit run with noise model
```

rum1887:
Do other AWS backends support noise inputs from Qiskit?

ruijterj:
Hi , can I work on this issue ?

christianbmadsen:
> Do other AWS backends support noise model input from Qiskit?

No. Today, Braket backends do not support Qiksit noise models. The Braket noise simulators/backends are [`braket_dm`](https://docs.aws.amazon.com/braket/latest/developerguide/braket-devices.html#braket-simulator-dm) and [`DM1`](https://docs.aws.amazon.com/braket/latest/developerguide/braket-devices.html#braket-simulator-dm1).

@rum1887 did you already start work on this issue? If not we will assign to @ruijterj.

rum1887:
> > Do other AWS backends support noise model input from Qiskit?
> 
[...]

PR #160, opened 2024-02-23
feat: support Braket noise model

yitchen-tim:
### Summary
Support Braket noise model for Braket backend. Solves half of https://github.com/qiskit-community/qiskit-braket-provider/issues/89. Still need to add support to Qiskit noise model which can be a separate PR.

### Details and comments
-  Support noise model in local dm simulator
```
local_dm_simulator = BraketLocalBackend(name='braket_dm', noise_model=noise_model)
```
- Support noise model in DM1 AWS backend
```
qpu_backend = provider.get_backend("dm1", noise_model=noise_model)
```
- This PR depends on https://github.com/amazon-braket/amazon-braket-sdk-python/pull/893. So checks will fail before this PR is merged.

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -10,6 +10,7 @@ from typing import Optional, Union
 from braket.aws import AwsDevice, AwsQuantumTask, AwsQuantumTaskBatch
 from braket.aws.queue_information import QueueDepthInfo
 from braket.circuits import Circuit
+from braket.circuits.noise_model import NoiseModel as BraketNoiseModel
 from braket.device_schema import DeviceActionType
 from braket.devices import Device, LocalSimulator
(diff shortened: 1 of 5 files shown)


yitchen-tim:
@speller26 I tried to implement the UX
```
dm1 = provider.get_backend("DM1", noise_model=noise_bit_flip)
```
One concern I have is the `provider.get_backend` depends on `provider.backends`. The args of `provider.backends` are for filtering. Now I added a `noise_model` arg which is not for filtering. It may make the `provider.backends` less cohesive, especially for user who use `backends` instead of `get_backend`. For example,
```
provider.backends(types="Simulator", noise_model=noise_bit_flip)
```
will try to add the noise model to all the returned backends. Looking for some feedback around this point.

yitchen-tim:
Closing this PR. The Braket backend in Qiskit should accept Qiskit noise model instead of Braket noise model.

PR #165, opened 2024-03-06
Deprecate AWSBraket_ for standardized classes

speller26:
Deprecates AWSBraketProvider and AWSBraketBackend for BraketProvider and BraketAwsBackend respectively


speller26:
Will update notebooks

Issue #130, opened 2023-11-10
Braket Provider returns measurement counts on all qubits instead of measured qubits


licedric:
The Qiskit-Braket provider returns measurement counts on all qubits, not just measured ones. This behavior is different from other Qiskit providers.

### Steps to reproduce the problem

```
sim = BraketLocalBackend()
qc = QuantumCircuit(2, 1)
qc.h(0)
qc.h(1)
qc.measure(0, 0)
task = execute(qc, sim)
result = task.result()
counts = result.get_counts()
print(counts)
```

### What is the current behavior?

Measurement results on all qubits, even those not explicitly measured are returned:
`{'11': 273, '00': 251, '10': 264, '01': 236}`

### What is the expected behavior?

Only results from the measured qubit are returned. This would match the behavior of other Qiskit backends.

licedric:
Counts are read in [this line](https://github.com/qiskit-community/qiskit-braket-provider/blob/ec7d04718550c7454c06922859ba0d6a62e4cd7b/qiskit_braket_provider/providers/braket_job.py#L51). Since Braket always measures all bits under the hood, this line will pull measurement counts on all qubits.

kshitijc:
The provider implements handling for measure [here](https://github.com/qiskit-community/qiskit-braket-provider/blob/ec7d04718550c7454c06922859ba0d6a62e4cd7b/qiskit_braket_provider/providers/adapter.py#L429), but it needs to access the sample result type [here](https://github.com/qiskit-community/qiskit-braket-provider/blob/ec7d04718550c7454c06922859ba0d6a62e4cd7b/qiskit_braket_provider/providers/braket_job.py#L51).

mdpratt:
Do we want to consider making an equivalent to qiskit's  `qc.measure_all()`?

ashlhans:
A fix for this issue is in progress [here](https://github.com/qiskit-community/qiskit-braket-provider/pull/169). This uses Braket's `measure` instruction and will measure the subset of qubits on any device that supports partial measurement.

speller26:
Completed in #169

PR #169, opened 2024-04-01
feature: add support for braket measure instruction

ashlhans:
### Summary

Add support for Braket's `measure` instruction.

This solves the bug detailed in [Issue #130 ](https://github.com/qiskit-community/qiskit-braket-provider/issues/130) where translating a circuit in qiskit to braket returned all the measurements instead of the requested measured qubit.

### Details and comments

Now, users can measure a subset of qubits on local simulators or any device that supports partial measurement.

Example:
```
sim = BraketLocalBackend()
qc = QuantumCircuit(2, 1)
qc.h(0)
qc.cx(0, 1)
qc.measure(0, 0)
task = transpile(qc, sim)
task = sim.run(task)
result = task.result()
counts = result.get_counts()
print(counts)
```
{'1': 518, '0': 506}

To keep the current functionality of the bug and measure all the qubits, simply change `measure(0, 0)` to `measure_all()`
```
sim = BraketLocalBackend()
qc = QuantumCircuit(2, 1)
qc.h(0)
qc.cx(0, 1)
qc.measure_all()
task = transpile(qc, sim)
task = sim.run(task)
result = task.result()
counts = result.get_counts()
print(counts)
```
{'00': 503, '11': 521}

## List of changes
- Update the amazon-braket-sdk version
- add `measure` to the Braket gates
- update adapter.py to use Braket's `measure`
- add tests

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -6,6 +6,7 @@ from math import inf, pi
 from typing import Optional, Union
 
 import braket.circuits.gates as braket_gates
+import braket.circuits.measure as measure
 import qiskit.circuit.library as qiskit_gates
 from braket.aws import AwsDevice
 from braket.circuits import (
@@ -66,6 +67,7 @@ _BRAKET_TO_QISKIT_NAMES = {
     "gpi2": "gpi2",
     "ms": "ms",
     "gphase": _GPHASE_GATE_NAME,
+    "measure": "measure"
 }
 
 _CONTROLLED_GATES_BY_QUBIT_COUNT = {
@@ -128,6 +130,7 @@ _GATE_NAME_TO_BRAKET_GATE: dict[str, Callable] = {
     "zz": lambda angle: [braket_gates.ZZ(2 * pi * angle)],
     # Global phase
     _GPHASE_GATE_NAME: lambda phase: [braket_gates.GPhase(phase)],
+    "measure": lambda: [measure.Measure()]
 }
 
 _QISKIT_CONTROLLED_GATE_NAMES_TO_BRAKET_GATES: dict[str, Callable] = {
@@ -182,6 +185,7 @@ _GATE_NAME_TO_QISKIT_GATE: dict[str, Optional[QiskitInstruction]] = {
         Parameter("theta") / (2 * pi),
     ),
     "gphase": qiskit_gates.GlobalPhaseGate(Parameter("theta")),
+    "measure": qiskit_gates.Measure()
 }
 
 
@@ -432,12 +436,7 @@ def to_braket(
         if gate_name == "measure":
             qubit = qubits[0]  # qubit count = 1 for measure
             qubit_index = circuit.find_bit(qubit).index
-            braket_circuit.sample(
-                observable=observables.Z(),
-                target=[
-                    qubit_index,
-                ],
-            )
+            braket_circuit.measure(qubit_index)
         elif gate_name == "barrier":
             warnings.warn(
                 "The Qiskit circuit contains barrier instructions that are ignored."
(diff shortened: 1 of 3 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -531,14 +529,17 @@ def to_qiskit(circuit: Circuit) -> QuantumCircuit:
     if not isinstance(circuit, Circuit):
         raise TypeError(f"Expected a Circuit, got {type(circuit)} instead.")
 
-    qiskit_circuit = QuantumCircuit(circuit.qubit_count)
+    num_measurements = sum(
+        instr.operator.name.lower() == "measure" for instr in circuit.instructions
```
speller26, line 533: Why a name check rather than `isinstance(instr.operator, Measure)`?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -128,6 +129,7 @@
     "zz": lambda angle: [braket_gates.ZZ(2 * pi * angle)],
     # Global phase
     _GPHASE_GATE_NAME: lambda phase: [braket_gates.GPhase(phase)],
+    "measure": lambda: [measure.Measure()],
```
speller26, line 132: Is this necessary? This dict is only used for gates.
```suggestion
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -66,6 +66,7 @@
     "gpi2": "gpi2",
     "ms": "ms",
     "gphase": _GPHASE_GATE_NAME,
+    "measure": "measure",
```
speller26, line 69: This isn't necessary because the dict is only used for transpilation
```suggestion
```

speller26 (commented):
Looking great!

qiskit_braket_provider/providers/adapter.py
ashlhans: This is used to match the qiskit `measure` to the Braket `Measure` so without it, the Braket measure isn't added to the Circuit. 

I do agree that this might not work well in this particular dict. Maybe there can be a `_ADDITIONAL_INSTRUCTION_NAME_TO_BRAKET_NAME` dict for things like measure?

qiskit_braket_provider/providers/adapter.py
ashlhans: Good point! An isinstance check is probably a safer way to go. Updated to use `isinstance(instr.operator, Measure)` instead.

qiskit_braket_provider/providers/adapter.py
speller26: That should be handled here: https://github.com/qiskit-community/qiskit-braket-provider/blob/0751b4c33f8f4cf262e273064eb241eb5e959a9d/qiskit_braket_provider/providers/adapter.py#L434-L437

ashlhans: Gotcha, that does work for the Qiskit to Braket conversion but the Braket to Qiskit conversion [here](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/qiskit_braket_provider/providers/adapter.py#L571-L573) checks for gates in the _GATE_NAME_TO_QISKIT_GATE dict. 

Is there a better way to implement the Braket to Qiskit conversion for measure?

speller26: Yeah, it's fine to leave it in the Braket->Qiskit direction; you could do a special check for measure operators, but that's not necessary.

ashlhans: Removed the `measure` gate from `_GATE_NAME_TO_BRAKET_GATE` but left it in `_GATE_NAME_TO_QISKIT_GATE`

speller26 (approved):

PR #180, opened 2024-06-06
Add noise model

### LINKED ISSUE #89: Support noise models

WingCode:
### Summary
closes #89

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -76,7 +76,9 @@ class BraketLocalBackend(BraketBackend):
         """
         super().__init__(name=name, **fields)
         self.backend_name = name
-        self._local_device = LocalSimulator(backend=self.backend_name)
+        self._local_device = LocalSimulator(
+            backend=self.backend_name, noise_model=fields.get("noise_model")
+        )
         self._target = local_simulator_to_target(self._local_device)
         self.status = self._local_device.status
 
--- a/qiskit_braket_provider/providers/braket_provider.py
+++ b/qiskit_braket_provider/providers/braket_provider.py
@@ -3,6 +3,7 @@
 import warnings
 
 from braket.aws import AwsDevice
+from braket.circuits.noise_model import NoiseModel
 from braket.device_schema.dwave import DwaveDeviceCapabilities
 from braket.device_schema.quera import QueraDeviceCapabilities
 from braket.device_schema.xanadu import XanaduDeviceCapabilities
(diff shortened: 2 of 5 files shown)


yitchen-tim:
Hi @WingCode, thanks for the contribution! This `qiskit-braket-provider ` package is designed for using Qiskit as frontend and Braket as backend. Therefore, the user experience is that user would create a _Qiskit_ circuit and [_Qiskit_ noise model](https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.noise.NoiseModel.html) to run on a Braket backend. Users do not pass _Braket_ noise model from the frontend. So, not only does the Qiskit noise model needs to propagate to the backend, the noise operations or the noise model need to be translated in some way. Because the Braket noise model and the Qiskit noise model does not support the same set of noise operations, some conversion may be needed based on physics. 

The issue https://github.com/qiskit-community/qiskit-braket-provider/issues/89 is about support Qiskit noise model with Braket backends in QIskit. 

The list of noises that the Braket local density matrix simulator supports can be found [here](https://github.com/amazon-braket/amazon-braket-default-simulator-python/blob/a4d7f98cb123ae6a1092972e728d2dbb93cb27b5/src/braket/default_simulator/density_matrix_simulator.py#L144). The definition of Braket noise operations can be found through [this mapping](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/d5dfbf40182f1858a2cf5db56962daf2e5e57cc3/src/braket/circuits/translations.py#L87).

PR #187, opened 2024-08-01
Move measurements after verbatim box

speller26:

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -436,6 +436,7 @@ def to_braket(
     _validate_name_conflicts(circuit.parameters)
 
     # Handle qiskit to braket conversion
+    measured_qubits = set()
     for circuit_instruction in circuit.data:
         operation = circuit_instruction.operation
         gate_name = operation.name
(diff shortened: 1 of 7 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -29,10 +29,12 @@
 from qiskit.circuit import Measure, Parameter, ParameterExpression
 from qiskit.circuit.parametervector import ParameterVectorElement
 from qiskit.transpiler import Target
-from qiskit_ionq import ionq_gates
+from qiskit_ionq import add_equivalences, ionq_gates
 
 from qiskit_braket_provider.exception import QiskitBraketException
 
+add_equivalences()
```
rmshaffer, line 36: It feels weird to run code like this at script import time. In the IonQ provider they do this as part of their backend `__init__` function: https://github.com/qiskit-community/qiskit-ionq/blob/f99c01f55dd74340216645abe25615a5256be0d7/qiskit_ionq/ionq_backend.py#L137

Is there a reason we cannot do something like that?

speller26: Figured I didn't want to re-run code every timeI instantiate the device, but I don't have strong feelings here

speller26: Thinking about it a bit more, I think it actually makes sense to run on import rather than instantiation (the way IonQ does it). If it's done on instantiation, then customers won't be able to use `to_braket` by itself to transpile to IonQ gates; they'd first need to instantiate a backend. In that case, it might make more sense to add the equivalences in `to_braket`; however, this is unnecessary extra work, as it only needs to happen once.

PR #189, opened 2024-08-14
Add native switch to run method

shpface:
- [x] I have added the tests to cover my changes.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.

### Summary
This change will add a `native=[bool]` switch to the `run()` method. This switch will ensure that the `to_braket` transpilation step uses:
* the device's native gate set
* the device's native topology

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -23,7 +23,8 @@ from .adapter import (
     aws_device_to_target,
     gateset_from_properties,
     local_simulator_to_target,
-    native_gate_support,
+    native_gate_connectivity,
+    native_gate_set,
     to_braket,
 )
 from .braket_quantum_task import BraketQuantumTask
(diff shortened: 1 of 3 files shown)


qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -327,8 +329,21 @@ def run(self, run_input, **options):
             del options["meas_level"]
 
         verbatim = options.pop("verbatim", False)
-        gateset = self._get_gateset() if not verbatim else None
-        braket_circuits = [to_braket(circ, gateset, verbatim) for circ in circuits]
+        native = options.pop("native", False)
```
speller26, line 332: Up to you whether you want to make this change, but it might be better to include these two arguments in the method signature instead of just popping options

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -327,8 +329,21 @@ def run(self, run_input, **options):
...
+        gateset = self._get_gateset()
```
speller26, line 333: Might be worth adding an option to `_get_gateset` to get the native gateset, so you don't need to get the non-native gateset each time

shpface: I think I like how that looks, thanks

speller26 (approved):

PR #196, opened 2025-02-28
Fix backend name

chongyouquan:


rmshaffer (approved):

rmshaffer (changes requested):
Code change makes sense, but the test is failing. Also lint is failing (which you can fix by running `black .` from the root of your repo and committing the resulting change).

rmshaffer:
CI tests failing because #197 is not yet merged; will update this branch and re-run CI once it is.

PR #202, opened 2025-05-28
Adding New Tutorial `native_programming`

### LINKED ISSUE #201: Add example for native programming
### What is the expected enhancement?
Add example notebook to the doc/tutorials folder for native programming, including 
1.  Visualize coupling map of a Braket device
2. View the native gates of a Braket device by `backend.get_gateset(native=True)` (change the existing private method `_get_gateset` to public)
3.  Example of programing a circuit with native gates that respect the device's coupling map. An non-trial example is preferred.
4. Run the circuit with verbatim mode, using `backend.run(circuit, verbatim=True)`

Reference: https://github.com/amazon-braket/amazon-braket-examples/blob/main/examples/braket_features/Verbatim_Compilation.ipynb

divshacker:
### Summary
fixes #201 

### Details and comments
All the details are mentioned in the above linked issue.


divshacker:
Hey @yitchen-tim, Can you please review this notebook? 
Please let me know the suggestions, I know there will be as this tutorial is still work in progress. 

Thanks

yitchen-tim:
Hi @divshacker thanks for the quick turn around! The explanation is great. For the coupling map, I didn't realize the IonQ's map becomes so messy. So, yes, let's keep the original plotting method. 
- Coupling map plot: The code for plotting seems repetitive. Can we define a `plot_coupling_map` function on top and reuse it?
- `_get_gateset`: Private methods are not supposed be called in application. Previously, there is no use case for the method, so the method was private. Now, we find a use case, and it's time to make that method public. So in addition to change in notebook, you also change the code base from `_get_gateset` to `get_gateset`, and update the places that call this method

yitchen-tim:
@divshacker Looks great! I there are couple style related issues. You can check it by running `tox` locally. I tried to fixed some but there are more than what I expected. Could you fix those style related issues so that checks can pass? 

After that, I will give it a pass, running the notebooks on QPU and we then are good to go!

divshacker:
Copy that

divshacker:
Hey @yitchen-tim , I tried to run `tox` locally but it is not showing errors, I never used tox btw 😅
Is it possible if you can just copy the style related issues as output by `tox` and paste it here in this PR. 
I will fix them by looking at the output.

divshacker:
Hey @yitchen-tim I tried to work around things here, 
I looked into the failed test and somehow found the style errors in the new tutorial. I change the tutorial accordingly. 
Let's see if the tests pass.

yitchen-tim:
Hi @divshacker , The checks still fail. The GitHub rule of this repo requires maintainers to approve the checks here which is a bit efficient. `tox` is a tool to run these checks locally. So if `tox` passes, the check here will pass.  If you can get `tox` set up locally, it will be more efficient. I think most highly maintained software projects uses `tox` or a similar tool. 

To set up `tox`, you can install the requirements in [requirements-dev.txt](https://github.com/qiskit-community/qiskit-braket-provider/blob/main/requirements-dev.txt), by running `pip install -r requirements-dev.txt`. If you run into problem about Pandoc (which I did), you can install the Pandoc on your computer. For me, I do `brew install pandoc` on my macbook. If you run on Windows or Linux machine, you may need a different command.

divshacker:
Hey @yitchen-tim , Can you now start test ? 

I have done some testing locally and fixed few changes accordingly.  
Let's see how it goes.

PR #203, opened 2025-05-31
Fix #200: Add angle restrictions support for Rigetti and IonQ backend…

### LINKED ISSUE #200: Extend `to_braket` to respect gate angle restrictions of native gates
Follow issue https://github.com/qiskit-community/qiskit-braket-provider/issues/138, the `to_braket` function converts the input QIskit circuit into a Braket circuit that contains the backend's native gates. However, such conversion does not range of angle is also not considered by the `to_braket` function.  The output circuit of  `to_braket` may currently contain angle not supported by the backend.

Specifically this issue should focus on 
* Rigetti (fixed angle list) and 
* IonQ (MS arbitrary angle range, see angle 3)

When apply the `to_braket` function to a QIskit circuit, shown as the code snippet below,
```
backend = AWSBraketProvider().get_backend("...")
circuit = QuantumCircuit(2)
circuit.x(0)
circuit.y(1)
circuit.rzz(1, 0, 1)
braket_circuit = to_braket(circuit, backend, native=True)
```
the output `braket_circuit` is a circuit in Braket Circuit class. The output Braket circuit should be equivalent to the input Qiskit circuit, but it only contains native gates of the device with the correct range of the angles. The two-qubit native gates apply only on the physical edge of the device.

### Acceptance criteria
- The output Braket circuit only use angles supported by the backend. Currently, the only backends that has such restriction in angles are Rigetti backends. As a native gate in Rigetti backends, the Rx can only accept +pi, -pi, +pi/2 and -pi/2.
- There is a extensible way to specify angle restriction for a backend. For example, a new restriction can be added to a backend without complex change in the code.

Note: A similar issue participated in UnitaryHack2024: https://github.com/qiskit-community/qiskit-braket-provider/issues/176

idriss-hamadi:
…s  - Add angle_restrictions.py with extensible restriction system 
- Modify to_braket() to accept backend parameter and respect angle limits 
- Implement RX decomposition for non-compliant angles on Rigetti devices 
- Add comprehensive test suite validating functionality 
- Maintain backward compatibility (backend parameter optional)  Resolves issue where to_braket output contained angles not supported by hardware.

### Summary

Added :
Complete angle_restrictions.py file
backend parameter to to_braket()
_op_to_instruction() function
Import statements
Restriction checking logic

Removed (~20 lines):
Inline gate conversion code (replaced with function call)
Direct gate creation (now goes through restriction checking)

Modified (~10 lines):
Function signatures
Main conversion loop
Parameter handling

### Details and comments

added some test file to assure the modification is working, can be deleted after review 

can update the documentation when receiving the review 
mainly the approach was : 

Detect when a backend has angle restrictions
Check if gate angles are allowed
Decompose disallowed angles into sequences of allowed operations
Make it extensible so new backends can be easily added

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -39,6 +39,11 @@ from qiskit.transpiler import Target
 from qiskit_ionq import add_equivalences, ionq_gates
 
 from qiskit_braket_provider.exception import QiskitBraketException
+from qiskit_braket_provider.providers.angle_restrictions import (
+    get_angle_restrictions,
+    is_angle_allowed,
+    decompose_rx_rigetti
+)
 
 add_equivalences()
 
@@ -464,6 +469,7 @@ def to_braket(
     basis_gates: Optional[Iterable[str]] = None,
     verbatim: bool = False,
     connectivity: Optional[list[list[int]]] = None,
+    backend=None,
 ) -> Circuit:
     """Return a Braket quantum circuit from a Qiskit quantum circuit.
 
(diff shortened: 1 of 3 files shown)


peterkomar-aws:
#200 is resolved. I'm closing this PR. Thank you for contributing.

PR #207, opened 2025-06-12
Recognize Kraus operators in to_braket and to_qiskit; support for common Braket noises

sesmart:
### Summary

Added support for converting Kraus operators between Qiskit and Braket. Qiskit has Kraus operator classes (qiskit.quantum_info.Kraus), that use sets of operators as data and can be fed to the braket.circuits.noises.Kraus class. Likewise, Braket SDK has Kraus operator instructions, as well as other common Noise instructions, which can be interpreted and applied as Kraus operators. These are added and tested in providers/adapter.py and providers/test_adapter.py, respectively. 

### Details and comments

/providers/adapter.py
- Added _BRAKET_SUPPORTED_ERRORS, which lists supported noise instructions from braket side
- to_braket: added kraus as handled case, converts to Braket Kraus object 
- to_qiskit; check if in supported errors, and if so run _create_qiskit_kraus, which treats the Kraus matrices as the parameters and feeds them into qiskit.quantum_info.Kraus 
- Support for noise operation through qiskit_aer somewhat supported, as these are realized as Kraus operators

/tests/providers/test_adapter.py
- test_kraus_to_braket ; test modified to_braket function 
- test_kraus_to_qiskit ; test modified to_qiskit function
- test_braket_noises ; check all braket noises can be converted to qiskit circuits
- test_kraus_compilation_q_to_b ; checking proper bit ordering, qiskit to braket
- test_kraus_compilation_b_to_q ; checking proper bit ordering, braket to qiskit 

On qiskit_aer noises: 
- Common qiskit noises are in the qiskit_aer module, where they have a separate QuantumError class. The QuantumError instruction creates a QuantumErrorInstruction, whose .definition is built using qiskit.quantum_info.Kraus. 
- These instructions are specifically named "quantum_channel", and are not caught by kraus type. Originally a workaround was tested, but it is difficult to build a suitable test without qiskit_aer. 
- Also found that `to_braket` will transpile them to standard qiskit Kraus operators, unless verbatim is specified. Verbatim on braket with Kraus operator may also have challenges with pragma, so this seems okay. quantum_channel / kraus seems likely to violate the `needs_transpilation` flag (line 543), so outside of verbatim, it should be transpiled. 

As another note, for two-qubit operations, in to_qiskit we reverse the matrices, and in to_braket we swap the qubit orderings.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -786,9 +786,8 @@ def _create_qiskit_kraus(gate_params: list[np.ndarray]) -> Instruction:
     """create qiskit.quantum_info.Kraus from Braket Kraus operators and reorder axes"""
     for n, param in enumerate(gate_params):
         assert (
-            (param.shape[0] == param.shape[1]),
-            "Kraus operators must be square matrices.",
-        )
+            param.shape[0] == param.shape[1]
+        ), "Kraus operators must be square matrices."
 
         n_q = int(np.log2(param.shape[0]))
         if n_q > 1:


qiskit_braket_provider/providers/adapter.py
```diff
@@ -197,6 +201,23 @@
     ),
     "gphase": qiskit_gates.GlobalPhaseGate(Parameter("theta")),
     "measure": qiskit_gates.Measure(),
+    "kraus": qiskit_qi.Kraus([[[1, 0], [0, 0]], [[0, 0], [0, 1]]]).to_instruction(),
+}
+
+_BRAKET_SUPPORTED_ERRORS = {
```
yitchen-tim, line 207: It appears that you only use the keys of this in the source code. `_BRAKET_SUPPORTED_ERRORS` can just be a list of names. The values are not needed i the source code. 

The values are only used in tests. You can add then as test data.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -197,6 +201,23 @@
     ),
     "gphase": qiskit_gates.GlobalPhaseGate(Parameter("theta")),
     "measure": qiskit_gates.Measure(),
+    "kraus": qiskit_qi.Kraus([[[1, 0], [0, 0]], [[0, 0], [0, 1]]]).to_instruction(),
```
yitchen-tim, line 204: This does not look right. `kraus` can be customized with Kraus operators, not just the depolarizing channel

qiskit_braket_provider/providers/adapter.py
```diff
@@ -6,7 +6,10 @@
 from typing import Optional, Union
 
 import braket.circuits.gates as braket_gates
+import braket.circuits.noises as braket_noisy
```
yitchen-tim, line 9: nit: let's use a consistent name ("braket_noises"). A similar comment below

qiskit_braket_provider/providers/adapter.py
yitchen-tim: nit: let's use a consistent name ("_BRAKET_SUPPORTED_NOISES").

qiskit_braket_provider/providers/adapter.py
```diff
@@ -751,6 +782,14 @@ def to_qiskit(circuit: Circuit) -> QuantumCircuit:
     return qiskit_circuit
 
 
+def _create_qiskit_kraus(gate_params: list[np.ndarray]) -> Instruction:
+    for param in gate_params:
```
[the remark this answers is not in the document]
yitchen-tim: I see, I overlooked that Braket only supports 2Q Kraus operators. But still good to add a validation in case BDK support more than 2Q Kraus in the future (in which case the logic here will break)

PR #218, opened 2025-08-26
feat: Support Qiskit 2.x

speller26:


speller26:
`test_kraus_braket_bit_ordering`  _sometimes_ fails with the error
> ValueError: cannot reshape array of size 16 into shape (2,2,2,2,2,2,2,2)

because `transpile` seems to be _nondeterministic_, coupled with the fact that the single-matrix Kraus instruction is converted to a UnitaryGate.

PR #223, opened 2025-09-12
fix: modify parameter assignment procedure + optional measurements

### LINKED ISSUE #222: Unnecessary conversion of float to FreeParameter

```
from qiskit import QuantumCircuit
from qiskit_braket_provider import to_qiskit, to_braket

qc = QuantumCircuit(1,1)
qc.rz(0.1, 0)
circ = Circuit().rz(0,0.1)

bqc1 = to_qiskit(to_braket(qc)) # passes 
bqc2 = to_qiskit(to_braket(to_qiskit(circ))) # fails
```

Running this following snippet 
```
circ = Circuit().rz(0,0.1)
print(type(circ.instructions[0].operator.angle))
print(type(to_braket(to_qiskit(circ)).instructions[0].operator.angle))
```
reveals the float has been converted to braket.parametric.free_parameter_expression.FreeParameterExpression. This raises an error in the to_qiskit, whereas supplying a FreeParameter alone does not. 

Circuits should be as close to fully interconvertable as possible.

sesmart:
### Summary

Closes #222, allowing for preservation of ParameterExpressions and FreeParameterExpressions in the conversion. This allows for a circuit to be converted from to_qiskit and back via to_braket. An added bonus is the option to not add measurements in the converter. 

### Details and comments

Problem: 
- Applying to_qiskit on a circuit with no FreeParameters (i.e simply floats) converted these to Parameter expressions with the values assigned to the float. Conversion back with to_braket recognized it as a Parameter or ParameterExpression and tried to create a FreeParameterExpression, but with the entire expression, not just the float value. This can lead to a few different types of errors. 

Solution:
- Added an option to not add measurement instructions in to_qiskit
- Directly feed gate_params into _create_qiskit_gate (no need to populate dict_param)
- Convert the FP or FPE to a ParameterExpression using sympy 
- Extract the coefficient from the existing template gate and combine it to form new parameters, and simply pass through numeric types
- Added tests

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -810,11 +810,11 @@ def _create_qiskit_gate(gate_name: str, gate_params: list[float | Parameter]) ->
         # extract the coefficient in the templated gate
         param = list(param_expression.parameters)[0].sympify()
         coeff = float(param_expression.sympify().subs(param, 1))
-        if hasattr(value, "expression"):
-            value = _sympy_to_qiskit(coeff * value.expression)
-        else:  # simply a numeric type
-            value = coeff * value
-        new_gate_params.append(value)
+        new_gate_params.append(
+            _sympy_to_qiskit(coeff * value.expression)
+            if hasattr(value, "expression")
+            else coeff * value
+        )
     return gate_cls(*new_gate_params)
 
 


qiskit_braket_provider/providers/adapter.py
```diff
@@ -801,10 +807,14 @@ def _create_qiskit_gate(gate_name: str, gate_params: list[float | Parameter]) ->
...
-        assigned = param_expression.assign(param, value)
-        new_gate_params.append(assigned)
+        # extract the coefficient in the templated gate
+        param = list(param_expression.parameters)[0].sympify()
+        coeff = float(param_expression.sympify().subs(param, 1))
+        if hasattr(value, "expression"):
+            value = _sympy_to_qiskit(coeff * value.expression)
+        else:  # simply a numeric type
+            value = coeff * value
+        new_gate_params.append(value)
```
speller26:
```suggestion
        new_gate_params.append(
            _sympy_to_qiskit(coeff * value.expression)
            if hasattr(value, "expression")
            else coeff * value
        )
```

PR #235, opened 2025-10-17
fix: Enable parallel transpilation

speller26:

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -532,7 +532,7 @@ def to_braket(
     Returns:
         Circuit | list[Circuit]: Braket circuit or circuits
     """
-    single_instance = isinstance(circuit, (Program, str)) or not isinstance(circuit, Iterable)
+    single_instance = isinstance(circuit, _Translatable) or not isinstance(circuit, Iterable)
     if single_instance:
         circuit = [circuit]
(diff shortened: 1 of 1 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -526,16 +524,26 @@ def to_braket(
...
+        num_processes (int | None): The maximum number of parallel transpilation processes for
+            multiple circuits. Default: `None`.
         pass_manager (PassManager): `PassManager` to transpile the circuit; will raise an error if
             used in conjunction with a target, basis gates, or connectivity. Default: `None`.
 
     Returns:
-        Circuit: Braket circuit
+        Circuit | list[Circuit]: Braket circuit or circuits
     """
     if isinstance(circuit, (Circuit, Program, str)):
```
sesmart: Would this now fail on a iterable of Circuits? From the input hinting translatable | iterable[translatable] that would seem to be an okay input.

tests/providers/test_adapter.py
sesmart: Can we add a test here for parallel transpilation?

qiskit_braket_provider/providers/adapter.py
speller26: Good catch, thanks!

sesmart (approved):

PR #241, opened 2025-10-23
fix: added check for multiples of m*pi/2

### LINKED ISSUE #233: Validation error for verbatim circuit for Ankaa-3

Run the following code
```
import json
from qiskit_braket_provider import to_braket
from qiskit import transpile, QuantumCircuit
from braket.aws import AwsDevice
qd = AwsDevice("arn:aws:braket:us-west-1::device/qpu/rigetti/Ankaa-3")

def ghz_pair(i,j):
    qc = QuantumCircuit(84)
    qc.h(i)
    qc.cx(i,j)
    bqc = transpile(qc, basis_gates=["rx","rz","iswap"],optimization_level=1)
    return to_braket(bqc,basis_gates=["rx","rz","iswap"])

test = ghz_pair(0,1)
res = qd.run(Circuit().add_verbatim_box(test),shots=100)
```

We hit a validation error saying

ValidationException: An error occurred (ValidationException) when calling the CreateQuantumTask operation: [line 6] Invalid RX angle '1.5707963267948968' with verbatim usage on Rigetti device. Valid angles are (-π, -π[/2](https://amazon-braket-device-emulator.notebook.us-east-1.sagemaker.aws/2), π[/2](https://amazon-braket-device-emulator.notebook.us-east-1.sagemaker.aws/2), π) or a custom gate can be declared via Pulse control.  If you are inputting float values of pi use a precision of 16

This is likely due to a precision provided from qiskit.

We expect the circuit to be run

sesmart:
### Summary

Close #233. Added a simple check for float parameter inputs to check if param is close to a rational $\frac{m}{2}\pi$, and replace it with a standard float.

### Details and comments

Found at least two cases where there is a precision output from the qiskit transpiler that should be validated okay on the service but leads to an error (on Rigetti, pi/2 multiple). Current check in to_braket should be very fast, and successfully catches test cases. 

Note, does not fix general issue of angle validation in the transpiler, which may be better addressed in another issue.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -774,9 +774,9 @@ def _create_free_parameters(operation):
                 renamed_param_name = _rename_param_vector_element(param)
                 params[i] = FreeParameterExpression(renamed_param_name)
             case param if isinstance(param, float):
-                mult = 12 * param / pi
+                mult = 2 * param / pi
                 if abs(mult - round(mult)) < _EPS:
-                    params[i] = round(mult) / 12 * pi
+                    params[i] = pi * round(mult) / 2
     return params
 
 


qiskit_braket_provider/providers/adapter.py
```diff
@@ -764,6 +764,10 @@ def _create_free_parameters(operation):
             case ParameterExpression():
                 renamed_param_name = _rename_param_vector_element(param)
                 params[i] = FreeParameterExpression(renamed_param_name)
+            case param if isinstance(param, float):
+                mult = 12 * param / pi
```
speller26: Why $\pi/12$ multiplies? Rigetti's only validation is for $\pi/2$ multiples

sesmart: That could make sense - I figured you could get to pi/4 as well, but it is true.

sesmart: Okay, restricted it to a pi/2 checks.

speller26 (approved):

PR #247, opened 2025-10-27
feat: Get allowed angles from gate calibrations

speller26:

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -727,6 +727,12 @@ def to_braket(
 ) -> Circuit | list[Circuit]:
     """Return a Braket quantum circuit from a Qiskit quantum circuit.
 
+    The recommended way to use this method is to minimally pass in a target (instead of basis gates
+    and connectivity) and qubit labels. The former ensures that the translated circuit is actually
+    supported by the device (and doesn't, for example, include unsupported parameters for gates).
+    The latter guarantees that the output Braket circuit uses the qubit labels of the Braket device,
+    which are not necessarily contiguous.
+
     Args:
         circuit (QuantumCircuit | Circuit | Program | str | Iterable): Qiskit or Braket circuit(s)
             or OpenQASM 3 program(s) to transpile and translate to Braket.


speller26:
Converting back to draft because Qiskit doesn't recognize the Ankaa gateset as universal with angles constrained
```python
to_braket(qc, target=backend.target)
```
```
TranspilerError: 'Unable to translate the operations in the circuit: {("h", 1), ("cx", 2), ("barrier", 4), ("measure", 1)} to the backend\'s (or manually specified) target basis: {"rx(3.141592653589793)", "rx(1.5707963267948966)", "rx(-3.141592653589793)", "rx(-1.5707963267948966)", "rz", "iswap", "measure", "barrier", "snapshot", "store"}. This likely means the target basis is not universal or there are additional equivalence rules needed in the EquivalenceLibrary being used. For more details on this error see: https://quantum.cloud.ibm.com/docs/api/qiskit/qiskit.transpiler.passes.BasisTranslator#translation-errors'
```
even though using the provider prior to this change produces gates with angles satisfying these constraints
```python
supported_angles = {3.141592653589793, 1.5707963267948966, -3.141592653589793, -1.5707963267948966}
[
    instruction for instruction in to_braket(qc, target=backend.target).instructions
    if instruction.operator.name.lower() == "rx" and instruction.operator.angle not in supported_angles
]
```
```
[]
```

speller26:
The custom pass works!

qiskit_braket_provider/providers/adapter.py
```diff
@@ -773,10 +937,6 @@ def _create_free_parameters(operation):
             case ParameterExpression():
                 renamed_param_name = _rename_param_vector_element(param)
                 params[i] = FreeParameterExpression(renamed_param_name)
-            case param if isinstance(param, float):
```
sesmart: Note the actual code this meant to fix now fails unfortunately - which was submitting without a target. Any harm in leaving this with a note that it addresses a particular edge case? 
```
    qc = QuantumCircuit(2)
    qc.h(0)
    qc.cx(0,1)
    nqc = to_braket(qc, basis_gates = ["rx","rz","iswap"], optimization_level = 1,
                    angle_restrictions={"rx":{0 : set([pi/2,pi, -pi/2, -pi])}})
    ... fails on Rigetti 
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -489,63 +531,183 @@ def _qpu_target(device: AwsDevice, description: str):
...
+    if parameter_restrictions:
+        _add_instructions_parameter_restrictions(
+            target,
+            parameter_restrictions,
+            instruction_props_1q,
+            instruction_props_2q,
+            default_props_2q,
+        )
+    else:
+        _add_instructions_no_parameter_restrictions(
```
sesmart: Seems like this could probably be brought into the same case where the no restrictions case returns the standard gateset?

tests/providers/test_braket_backend.py
```diff
@@ -569,6 +585,91 @@ def test_target(self):
...
+                Circuit().add_verbatim_box(
+                    Circuit().rz(0, np.pi / 2).rx(0, np.pi / 2).rz(0, np.pi / 2)
+                ),
+            )
+            target._pass_manager.run = lambda circuits: circuits
+            self.assertEqual(
+                to_braket(qc, target=target),
+                Circuit().add_verbatim_box(Circuit().rz(0, np.pi / 2).v(0).rz(0, np.pi / 2)),
+            )
+
```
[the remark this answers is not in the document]
speller26: I'd say that's an issue with Qiskit itself; that might just be the consequence of asking the transpiler to use {X, SX, SXdg} instead of arbitrary Rx.

sesmart (approved):

PR #250, opened 2025-11-11
feat: Braket-native primitives

speller26:
### Summary

Introduces `BraketEstimator` and `BraketSampler`.

`BraketEstimator` is a native Qiskit estimator for Amazon Braket that converts estimator PUBs into circuit bindings. It supports arbitrary shapes for parameters and observables by creating potentially multiple CircuitBindings for a single PUB; each Cartesian product of parameters with a Pauli sum or monomial array maps to a single CircuitBinding.

`BraketSampler` creates a single CircuitBinding for each sampler PUB; it supports any arbitrary registers in the circuit and packs bits according to Qiskit convention.

### DIFF
--- a/tests/providers/test_braket_sampler.py
+++ b/tests/providers/test_braket_sampler.py
@@ -51,10 +51,18 @@ class TestBraketSampler(TestCase):
             chsh_circuit.cx(i, i + 1)
         chsh_circuit.ry(theta, 0)
         chsh_circuit.measure_all(add_bits=False)
-        parameter_values = np.array(  # shape (3, 6)
-            [np.linspace(0, 2 * np.pi, 6), np.linspace(0, np.pi, 6), np.linspace(np.pi, 2 * np.pi, 6)]
+        pub = SamplerPub.coerce(
+            (
+                chsh_circuit,
+                np.array(  # shape (3, 6)
+                    [
+                        np.linspace(0, 2 * np.pi, 6),
+                        np.linspace(0, np.pi, 6),
+                        np.linspace(np.pi, 2 * np.pi, 6),
+                    ]
+                ),
+            )
         )
-        pub = SamplerPub.coerce((chsh_circuit, parameter_values))
 
         data = self.sampler.run([pub]).result()[0].data
         data_backend = BackendSamplerV2(backend=self.backend).run([pub]).result()[0].data


qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+                )
+            )
+
+            # Map each position in the broadcast to its location in the binding result
+            obs_idx_map = {ok: idx for idx, (ok, _) in enumerate(monomials)}
+            binding_to_result_map[len(bindings) - 1] = [
+                (position, obs_idx_map[ok], param_idx_map[pi])
+                for ok, _ in monomials
+                for position, pi in obs_groups[ok]
+            ]
```
rmshaffer: what if `monomials` is empty? do you still want to do all this?

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        data = defaultdict(list)
+        for bindings_array in param_list:
+            for k, v in bindings_array.data.items():
+                for param, val in zip(k, v):
+                    data[param].append(val)
+        return ParameterSets(data)
+
+    @staticmethod
+    def _translate_observables(obs_list: list) -> list[Observable]:
```
rmshaffer: this feels like a good function to have dedicated unit tests for. (perhaps also `_pauli_to_observable`)

tests/providers/test_braket_estimator.py
```diff
@@ -0,0 +1,396 @@
...
+            for observable_set in observables
+        ]
+        estimator_pub = circuit, observables, params
+
+        self.assertTrue(
+            np.allclose(
+                self.estimator.run([estimator_pub]).result()[0].data.evs,
+                BackendEstimatorV2(backend=self.backend).run([estimator_pub]).result()[0].data.evs,
+                rtol=0.3,
+                atol=0.2,
```
rmshaffer: why do these tolerances need to be so high? could we reduce them if we (for example) increase the number of shots?

same comment for the following test.

rmshaffer (commented):
Looks awesome! Couple of small comments. I didn't thoroughly review the test code.

It would be great if we could also add example notebooks for both primitives & transpilation in this repo before we push the v0.9 release, since that will (a) give users documentation, and (b) perhaps allow us to catch any lurking bugs in the end-to-end flow.

qiskit_braket_provider/providers/braket_sampler.py
```diff
@@ -0,0 +1,207 @@
...
+    ) -> BraketPrimitiveTask:
+        """
+        Samples circuits with multiple parameter values.
+        Args:
+            pubs (Iterable[SamplerPubLike]): An iterable of SamplerPubLike objects to sample.
+            shots (int): Number of shots to run for each circuit. Default: 1024
+        Returns:
+            BraketPrimitiveTask: A job object containing the sample results.
+        """
+        coerced_pubs = [SamplerPub.coerce(pub, shots) for pub in pubs]
```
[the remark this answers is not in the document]
speller26: Fixed this behavior (and that of precision in the estimator) by removing shots from `.coerce` and deciding on the shots value later.

speller26:
I'll add notebooks in a subsequent PR (and probably amazon-braket-examples too)

PR #262, opened 2025-12-05
Update older examples to use new primitives

speller26:


sesmart (approved):
Nice work! Looks good to me.

PR #265, opened 2025-12-06
fix: Support sampler PUBs with no parameters

speller26:


qiskit_braket_provider/providers/braket_sampler.py
speller26: The changes here (and test_braket_sampler.py) are the main logical change of this PR; everything in the other files is either extra tests or minor edits.

PR #274, opened 2026-01-20
fix: lowered opt level to avoid relabelling

sesmart:
### Summary

Failing [test_braket_backend::test_run_with_pass_manager](https://github.com/qiskit-community/qiskit-braket-provider/actions/runs/21157949424/job/60846513491). 

### Details and comments

As all 1Q error rates are equal, relabeling should not occur - might occur with some rng with opt level 2


sesmart:
Closed by https://github.com/qiskit-community/qiskit-braket-provider/pull/276.

PR #275, opened 2026-01-22
feat: support barrier

yitchen-tim:
### Summary
Implement translation of Qiskit barrier instructions to Braket barrier compiler directives. Previously barriers were ignored with a warning. Now they are preserved in Braket circuits.


nilsquet (dismissed):
Looks good to me. We could just add an edge case test for creating a Qiskit QuantumCircuit with `0` qubits and add a global barrier. 

Also, it's interesting to note that a global Qiskit barrier does never translate to a global Braket barrier, because there is always specific qubits assigned. This is slightly different to our barrier handling in the BDK (at least if https://github.com/amazon-braket/amazon-braket-sdk-python/pull/1203 is merged).

speller26 (approved):

PR #280, opened 2026-02-26
feat: add support for braket verbatim pragmas

aniksd-braket:
### Summary
- support braket verbatim pragmas when converting OQ3 programs using `to_qiskit()`
- support braket verbatim pragmas when compiling using `to_braket()`


qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +306,85 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+    This context extends AbstractProgramContext to build Qiskit QuantumCircuits from
+    OpenQASM 3 programs. It supports Braket verbatim pragmas, which are converted to
+    Qiskit BoxOp operations to preserve gate sequences that should not be optimized.
+    
+    Verbatim boxes are represented using Qiskit's native BoxOp construct, which treats
+    a block of operations atomically. All verbatim boxes in a circuit use the same
+    configurable label name.
+    
+    Args:
+        verbatim_box_name: Name to use for BoxOp labels when converting verbatim boxes.
```
nilsquet: The argument description is repeated below in the respective `init` method and should be removed here.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +306,85 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        """Initialize the Qiskit program context.
+        
+        Args:
+            verbatim_box_name: Name to use for BoxOp labels when converting verbatim boxes.
+                Default: "verbatim"
+        """
         super().__init__()
         self._circuit = QuantumCircuit()
         self._param_map = {}
+        # Verbatim pragma handling infrastructure
```
nilsquet: I suggest to remove the comment because each of the three variables has `verbatim` in it.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +306,85 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        """
+        super().declare_variable(name, symbol_type, value, const)
+        
+        # If this is a bit type declaration, add classical bits to the circuit
+        if isinstance(symbol_type, BitType):
+            # Get the size of the bit array
+            if symbol_type.size is not None:
+                # Extract the integer value from IntegerLiteral
+                # If size is an Identifier (e.g., function parameter), skip adding bits
+                if hasattr(symbol_type.size, 'value'):
```
nilsquet: Could we use something like `isinstance(symbol_type.size, IntegerLiteral)` instead to be sure `value` is an `int`?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +306,85 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                    size = symbol_type.size.value
+                else:
+                    # Size is an Identifier or expression, can't determine size yet
+                    # This happens for function parameters like bit[n] where n is a variable
+                    return
+            else:
+                size = 1
+            
+            # Add classical bits to the circuit
```
nilsquet: I would remove that comment because it doesn't add any more information.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +306,85 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            for _ in range(size):
+                self._circuit.add_bits([Clbit()])
```
nilsquet:
```suggestion
            self._circuit.add_bits([Clbit() for _ in range(size)])
```
This way `add_bits` is only called once.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
             gate = gate.control(
                 len(ctrl_modifiers), ctrl_state=str("".join([str(i) for i in ctrl_modifiers]))
             )
-        self._circuit.append(CircuitInstruction(gate, target))
+
+        # Ensure circuit has enough qubits for the target indices
+        # This is needed when using physical qubits ($0, $1, etc.) where no qubit register is declared
+        max_qubit_index = max(target) if target else -1
+        while self._circuit.num_qubits <= max_qubit_index:
```
nilsquet: Can we determine the number of needed qubits and call `add_bits` only once?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
...
+            self._circuit.add_bits([Qubit()])
+
+        # Redirect to verbatim circuit if inside a verbatim box
+        if self._in_verbatim_box:
+            # Ensure verbatim circuit also has enough qubits
+            while self._verbatim_circuit.num_qubits <= max_qubit_index:
+                self._verbatim_circuit.add_bits([Qubit()])
```
nilsquet: Same here.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
...
+            self._circuit.append(CircuitInstruction(gate, target))
 
     def handle_parameter_value(self, value: Number | Expr) -> Number | Parameter:
         return _sympy_to_qiskit(value, self._param_map) if isinstance(value, Expr) else value
 
-    def add_measure(self, target: tuple[int], classical_targets: Iterable[int] = None):
+    def add_measure(self, target: tuple[int], classical_targets: Iterable[int] | None = None):
         for iter, qubit in enumerate(target):
             index = classical_targets[iter] if classical_targets else iter
             self._circuit.measure(qubit, index)
```
nilsquet, line 432: This doesn't distinguish by `_in_verbatim_box`, right? Does it mean that any measurement in the verbatim box would end up outside the box?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
...
+        - Append the BoxOp to the main circuit
+        - Reset verbatim state
+
+        Args:
+            marker: VerbatimBoxDelimiter indicating START_VERBATIM or END_VERBATIM
+
+        Raises:
+            ValueError: If nested verbatim boxes are encountered or if END_VERBATIM is called without START_VERBATIM
+        """
+        from qiskit.circuit import BoxOp
```
nilsquet: All imports should be moved to the top.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
...
+
+        if marker == VerbatimBoxDelimiter.START_VERBATIM:
+            # Check for nested verbatim boxes
+            if self._in_verbatim_box:
+                raise ValueError("Nested verbatim boxes are not supported")
+
+            # Create a new circuit to collect verbatim gates
+            self._verbatim_circuit = QuantumCircuit(self.num_qubits)
```
nilsquet: What would happen if we have an entire verbatim-only circuit? Is `num_qubits` here available as well?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +407,76 @@ def add_gate_instruction(
...
+            self._in_verbatim_box = True
+
+        elif marker == VerbatimBoxDelimiter.END_VERBATIM:
```
nilsquet, line 460: We could add an else branch that raises an error.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +877,134 @@ def _add_instructions_no_parameter_restrictions(
...
+    """
+    from qiskit.circuit import Barrier, BoxOp
+
+    # Create new circuit with same registers
+    modified_circuit = QuantumCircuit(*circuit.qregs, *circuit.cregs)
+    modified_circuit.global_phase = circuit.global_phase
+
+    verbatim_boxes = []
+
+    # Iterate through all instructions
```
nilsquet: This method has many inline comments and not all of them are adding value (e.g., this one does not).

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +877,134 @@ def _add_instructions_no_parameter_restrictions(
...
+
+    Returns:
+        A tuple of (modified_circuit, verbatim_boxes) where:
+        - modified_circuit: Circuit with BoxOps replaced by named barriers
+        - verbatim_boxes: List of (box_circuit, qubit_indices) tuples
+    """
+    from qiskit.circuit import Barrier, BoxOp
+
+    # Create new circuit with same registers
+    modified_circuit = QuantumCircuit(*circuit.qregs, *circuit.cregs)
```
nilsquet:
```python
from qiskit_braket_provider.providers.adapter import to_qiskit

oq3 = '''OPENQASM 3.0;
h $0;
#pragma braket verbatim
box { cnot $0, $1; }
'''
qc = to_qiskit(oq3)
```

creates a `QuantumCircuit` with `qc.qregs==[]` which leads to an error when calling `qc2 = to_braket(qc)/to_braket(oq3)`. Is this expected?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +877,134 @@ def _add_instructions_no_parameter_restrictions(
...
+        verbatim_box_name: The label name used to identify verbatim barriers
+
+    Returns:
+        Circuit with verbatim box contents restored
+
+    Raises:
+        ValueError: If barrier count doesn't match verbatim box count
+        ValueError: If qubit mapping fails
+    """
+    from qiskit.circuit import Barrier
```
nilsquet: Import should be moved to the top.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +877,134 @@ def _add_instructions_no_parameter_restrictions(
...
+    for instruction in transpiled_circuit.data:
+        operation = instruction.operation
+
+        # Check if this is a barrier with the verbatim box name label
+        if isinstance(operation, Barrier) and getattr(operation, "label", None) == verbatim_box_name:
+            barrier_count += 1
+
+            # Get the next verbatim box
+            try:
+                box_circuit, _original_qubit_indices = next(verbatim_box_iter)
```
nilsquet: `_original_qubit_indices` is not used

qiskit_braket_provider/providers/adapter.py
```diff
@@ -832,6 +1103,28 @@ def to_braket(
     )
     coupling_map = coupling_map or connectivity
 
+    # Check if any circuits have verbatim boxes and extract them before transpilation
+    from qiskit.circuit import BoxOp
```
nilsquet: Import should be moved to the top.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -851,6 +1144,17 @@ def to_braket(
...
+        
+        # Determine transpilation parameters based on verbatim boxes
+        if has_verbatim_boxes:
+            # Use trivial layout and no routing unless user overrides
+            effective_layout_method = layout_method if layout_method is not None else 'trivial'
+            effective_routing_method = routing_method if routing_method is not None else 'none'
+        else:
+            # Use user-provided values or Qiskit defaults
+            effective_layout_method = layout_method
+            effective_routing_method = routing_method
```
nilsquet, line 1147: Adding these two parameters is not connected to the verbatim implementation, or am I overseeing something?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -851,6 +1144,17 @@ def to_braket(
     elif not verbatim:
         # If basis_gates is not None, then target remains empty
         target = target if basis_gates or target else _default_target(circuits)
+        
+        # Determine transpilation parameters based on verbatim boxes
+        if has_verbatim_boxes:
+            # Use trivial layout and no routing unless user overrides
+            effective_layout_method = layout_method if layout_method is not None else 'trivial'
+            effective_routing_method = routing_method if routing_method is not None else 'none'
+        else:
```
nilsquet: I thought it must be `trivial` for the layout and `None` for the routing if verbatim boxes are present?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1195,22 +1509,54 @@ def _translate_pauli(pauli: Pauli, coeff: float = 1.0) -> BraketObservable:
...
+    Examples:
+        Convert an OpenQASM 3 program with a verbatim box:
+        
+        >>> openqasm_program = '''
+        ... OPENQASM 3.0;
+        ... qubit[2] q;
+        ... #pragma braket verbatim
+        ... box {
+        ...     h q[0];
+        ...     cnot q[0], q[1];
```
nilsquet: Shouldn't verbatim circuits use physical qubits?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +877,134 @@ def _add_instructions_no_parameter_restrictions(
...
+    Args:
+        circuit: The Qiskit circuit to process
+        verbatim_box_name: The label name used to identify verbatim BoxOp operations
+
+    Returns:
+        A tuple of (modified_circuit, verbatim_boxes) where:
+        - modified_circuit: Circuit with BoxOps replaced by named barriers
+        - verbatim_boxes: List of (box_circuit, qubit_indices) tuples
+    """
+    from qiskit.circuit import Barrier, BoxOp
```
nilsquet: Should be moved to the top.

nilsquet (changes requested):
Thanks a lot for this PR, @aniksd-braket. It's very visible that this was a lot of work that lead to a solid implementation. I just went through the `adapter.py` in detail and added several comments. 

Since this PR comes with a high number of changed lines of code, it is quite time-consuming to review. I have a few suggestions how to reduce it to make it more digestible:
- There is a large number of unrelated formatting changes in files that are unrelated to the features you implement here. It would be beneficial to split out those changes (e.g., all files but the `adapter.py` and its related tests).
- All imports should be at the top of the file to save a few redundant import calls.
- There is a high number of inline comments, but not all of them are meaningful. We could remove some of them.
- The tests are very exhaustive, but a lot of boiler plate code is repeated. Using parameterized test would significantly reduce the number of lines of code here.

Regarding the tests itself, I noticed two things:
1) The tests use OQ3 programs with declared instead of physical qubits. Shouldn't we always use physical qubits in the verbatim box?
2) In `test_adapter_verbatim.py`, the assertions are very specific when checking the expected result circuit which I highly appreciate. However, in `test_adapter_to_braket_verbatim.py`, this is not the case and often `assert braket_circuit is not None` is used.

qiskit_braket_provider/providers/adapter.py
laurencap: +1, to me looks like the kind of comment an AI agent would generate (idk why they like to add lots of noisy comments, you can use a steering file to encourage kiro not to do this)

docs/tutorials/3_tutorial_minimum_eigen_optimizer.ipynb
```diff
@@ -18,19 +18,19 @@
   },
   {
    "cell_type": "code",
+   "execution_count": 1,
```
laurencap, line 21: is this file in scope for this change? I would be surprised if an existing tutorial was modified by a new feature

qiskit_braket_provider/providers/adapter.py
aniksd-braket: yup measurement are always outside the verbatim box. Correct me if I am wrong @laurencap

qiskit_braket_provider/providers/adapter.py
aniksd-braket: It is always initialized to 0. And they are added dynamically if required as we encounter operation on qubits.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Great find, fixing this now

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Yup, fixing this

qiskit_braket_provider/providers/adapter.py
aniksd-braket: It is related, as we realized we needed this only after adding support for verbatim pragmas. We decided to expand the support to give more compilation control for customers using qiskit circuits directly

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +307,78 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        if self._in_verbatim_box:
+            raise ValueError(
+                "Unclosed verbatim box at end of program. "
+                "Every verbatim box start marker must have a matching end marker."
+            )
         return self._circuit
 
     def add_qubits(self, name: str, num_qubits: int | None = 1) -> None:
         super().add_qubits(name, num_qubits)
         self._circuit.add_register(num_qubits, num_qubits)
```
nilsquet: Why are `num_qubits` many classical bits added?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +401,77 @@ def add_gate_instruction(
             gate = gate.control(
                 len(ctrl_modifiers), ctrl_state=str("".join([str(i) for i in ctrl_modifiers]))
             )
-        self._circuit.append(CircuitInstruction(gate, target))
+
+        # Ensure circuit has enough qubits for the target indices
+        # This is needed when using physical qubits ($0, $1, etc.) where no qubit register is declared
+        max_qubits = (max(target) + 1) if target else -1
+        self._circuit.add_bits([Qubit() for _ in range(max_qubits - self._circuit.num_qubits)])
+        self.num_qubits = self._circuit.num_qubits
```
nilsquet, line 412:
```suggestion
```
This is not used.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +401,77 @@ def add_gate_instruction(
             gate = gate.control(
                 len(ctrl_modifiers), ctrl_state=str("".join([str(i) for i in ctrl_modifiers]))
             )
-        self._circuit.append(CircuitInstruction(gate, target))
+
+        # Ensure circuit has enough qubits for the target indices
+        # This is needed when using physical qubits ($0, $1, etc.) where no qubit register is declared
+        max_qubits = (max(target) + 1) if target else -1
+        self._circuit.add_bits([Qubit() for _ in range(max_qubits - self._circuit.num_qubits)])
```
nilsquet: Can we move this into the respective if branch where it is needed?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +401,77 @@ def add_gate_instruction(
...
+
+            # Create BoxOp with the verbatim circuit
+            box_op = BoxOp(self._verbatim_circuit, label=self._verbatim_box_name)
+
+            # Append BoxOp to main circuit with all qubits (convert indices to Qubit objects)
+            # We need to pass the actual Qubit objects from the main circuit, not just indices
+            qubit_objects = [self._circuit.qubits[i] for i in range(self._verbatim_circuit.num_qubits)]
+            self._circuit.append(box_op, qubit_objects)
+
+            # Reset verbatim state
```
nilsquet: There is still a number of comments that don't add value and, therefore, should be removed.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -832,6 +1096,28 @@ def to_braket(
...
+
+    has_verbatim_boxes = any(
+        any(
+            isinstance(instr.operation, BoxOp)
+            and getattr(instr.operation, "label", None) == verbatim_box_name
+            for instr in circ.data
+        )
+        for circ in circuits
+    )
+
```
nilsquet: Should we add something like
```python
if pass_manager and has_verbatim_boxes:
    raise ValueError(
        "Custom pass_manager is not supported with verbatim boxes."
        "Verbatim boxes require controlled transpilation to preserve gate ordering."
    )
```

to avoid that people call, e.g., `RemoveBarriers()` in a custom pass manager?

qiskit_braket_provider/providers/adapter.py
nilsquet: I suggest to split off this added functionality into a separate CR. Also, we should give a warning that these parameters are overridden when using verbatim.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +872,132 @@ def _add_instructions_no_parameter_restrictions(
...
+                clbit_indices = [box_circuit.find_bit(q).index for q in box_instruction.clbits]
+                # Append the gate instruction with the same qubits as in the box
+                reconstructed_circuit.append(
+                    box_instruction.operation, qubit_indices, clbit_indices
+                )
+        else:
+            # Get indices of qubits and clbits and add instruction as-is
+            qubit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.qubits]
+            clbit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.clbits]
+            reconstructed_circuit.append(operation, instruction.qubits, instruction.clbits)
```
nilsquet: `qubit_indices` and `clbit_indices` are computed but `instruction.qubits, instruction.clbits` are used.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -260,6 +270,8 @@
     "Z": braket_observables.Z,
 }
 
+_BRAKET_VERBATIM_BOX_NAME = "verbatim"
```
nilsquet, line 278: Should we use a name that users will most likely not pick by accident? Because if they use this identifier for a regular barrier, the code breaks.

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,473 @@
...
+
+from qiskit_braket_provider.providers.adapter import (
+    _extract_verbatim_boxes,
+    _restore_verbatim_boxes,
+    to_braket,
+    to_qiskit,
+)
+
+
+# --- Helpers ---
```
nilsquet: Comments like this should be removed.

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,473 @@
...
+    def test_to_braket_with_non_verbatim_boxops(self):
+        inner = _make_box_circuit(2, [("h", [0]), ("cx", [0, 1])])
+        qc = QuantumCircuit(2)
+        qc.append(BoxOp(inner, label="other_label"), [0, 1])
+
+        try:
+            bc = to_braket(qc, verbatim=False)
+            assert bc is not None
+        except Exception:
+            pass  # transpiler may fail on non-verbatim BoxOp — expected
```
nilsquet: This test always passes and doesn't test anything.

tests/providers/test_adapter.py
```diff
@@ -260,7 +260,7 @@ def test_u_gate(self):
     def test_standard_gate_decomp(self):
```
nilsquet, line 260: The changes in this file seem unrelated and only concern formatting, right?

tests/providers/test_adapter_to_qiskit_verbatim.py
```diff
@@ -0,0 +1,306 @@
+"""Tests for verbatim pragma support in Qiskit to Braket adapter."""
```
nilsquet, line 1: Should we name this file "test_adapter_to_qiskit_verbatim.py" to be consistent?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +872,132 @@ def _add_instructions_no_parameter_restrictions(
...
+        ValueError: If qubit mapping fails
+    """
+    # Create new circuit with same registers
+    reconstructed_circuit = QuantumCircuit(transpiled_circuit.num_qubits, transpiled_circuit.num_clbits)
+    reconstructed_circuit.global_phase = transpiled_circuit.global_phase
+
+    # Create iterator over verbatim boxes
+    verbatim_box_iter = iter(verbatim_boxes)
+    barrier_count = 0
+
```
nilsquet, line 944: Instead of doing the try/except below, could we just do something like this:

```python
barrier_count = sum(
    1 for instr in transpiled_circuit.data
    if isinstance(instr.operation, Barrier) and getattr(instr.operation, "label", None) == verbatim_box_name
)
if barrier_count != len(verbatim_boxes):
    raise ValueError(...)
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +872,132 @@ def _add_instructions_no_parameter_restrictions(
...
+    for instruction in transpiled_circuit.data:
+        operation = instruction.operation
+
+        # Check if this is a barrier with the verbatim box name label
+        if isinstance(operation, Barrier) and getattr(operation, "label", None) == verbatim_box_name:
+            barrier_count += 1
+
+            # Get the next verbatim box
+            try:
+                box_circuit, _ = next(verbatim_box_iter)
```
nilsquet, line 952: Then, we don't need the try/except.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -741,12 +872,132 @@ def _add_instructions_no_parameter_restrictions(
...
+    # Create new circuit with same registers
+    reconstructed_circuit = QuantumCircuit(transpiled_circuit.num_qubits, transpiled_circuit.num_clbits)
+    reconstructed_circuit.global_phase = transpiled_circuit.global_phase
+
+    # Create iterator over verbatim boxes
+    verbatim_box_iter = iter(verbatim_boxes)
+    barrier_count = 0
+
+    # Iterate through all instructions in transpiled circuit
+    for instruction in transpiled_circuit.data:
```
nilsquet, line 945: Instead of iterating through the circuit and re-building it entirely, couldn't we just replace the barriers with the verbatim boxes? That way, we would eliminate the `else` branch and improve memory efficiency.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -832,6 +1096,28 @@ def to_braket(
...
+    # Check if any circuits have verbatim boxes and extract them before transpilation
+    from qiskit.circuit import BoxOp
+
+    has_verbatim_boxes = any(
+        any(
+            isinstance(instr.operation, BoxOp)
+            and getattr(instr.operation, "label", None) == verbatim_box_name
+            for instr in circ.data
+        )
+        for circ in circuits
```
nilsquet: This flag is global across the batch and if any circuit has a verbatim box, all circuits go through extract/restore. Could we instead only do it for the circuits that have verbatim boxes?

qiskit_braket_provider/providers/adapter.py
aniksd-braket: I agree with giving a warning, but I don't think this should be a separate CR. It is very much related to the contents of this PR which is providing verbatim support.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: This was the existing implementation. @speller26 might have a better idea of why this decision was made.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: True, but I am just updating the existing instance variable `self.num_qubits` to accurately represent the number of qubits in the circuit.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Nope unfortunately we cannot do that as Qiskit does not have non-hacky way of adding gates to the middle of a circuit

aniksd-braket: Nope, unfortunately we cannot do that as Qiskit does not have non-hacky way of adding gates to the middle of a circuit

qiskit_braket_provider/providers/adapter.py
aniksd-braket: It should be outside too, because the outer circuit also needs the extra qubits. This is useful when restoring the verbatim boxes back post-compilation

qiskit_braket_provider/providers/adapter.py
aniksd-braket: I would lean towards continue using this as it is the one used in the service. I can add an error message instead

tests/providers/test_adapter.py
aniksd-braket: Yes, that's correct, these were annoying to separate out. So left it in this PR

aniksd-braket: once the linter branch gets merged, this change will not be visible though

qiskit_braket_provider/providers/adapter.py
aniksd-braket: I would lean towards keeping the try/except as the approach you suggested iterates over `transpiled_circuit.data` twice, while the current approach does so only once. This makes a tiny bit more performant.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -850,6 +1160,18 @@ def to_braket(
     elif not verbatim:
         # If basis_gates is not None, then target remains empty
         target = target if basis_gates or target else _default_target(circuits)
+        
+        # Determine transpilation parameters based on verbatim boxes
+        if has_verbatim_boxes:
+            warnings.warn("Overrriding layout method to 'trivial' and routing method to 'none' as the circuit as verbatim blocks", stacklevel=1)
```
nilsquet:
```suggestion
            warnings.warn("Overriding layout method to 'trivial' and routing method to 'none' as the circuit has verbatim blocks", stacklevel=1)
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -831,6 +1102,45 @@ def to_braket(
...
+            isinstance(instr.operation, BoxOp)
+            and getattr(instr.operation, "label", None) == verbatim_box_name
+            for instr in circ.data
+        )
+        for circ in circuits
+    )
+    
+    if pass_manager and has_verbatim_boxes:
+        raise ValueError(
+            "Custom pass_manager is not supported with verbatim boxes."
```
nilsquet:
```suggestion
            "Custom pass_manager is not supported with verbatim boxes. 
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -740,6 +876,128 @@ def _add_instructions_no_parameter_restrictions(
...
+    Returns:
+        A tuple of (modified_circuit, verbatim_boxes) where:
+        - modified_circuit: Circuit with BoxOps replaced by named barriers
+        - verbatim_boxes: List of (box_circuit, qubit_indices) tuples
+    """
+
+    if not (any(isinstance(instr.operation, BoxOp)
+                and getattr(instr.operation, "label", None) == verbatim_box_name
+                for instr in circuit.data)):
+        return circuit, []
```
nilsquet: This is meant as an shortcut to return early, correct? Since it also iterates `circuit.data`, I don't think it has a significant benefit and would remove it.

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,470 @@
...
+class TestRoundTripConversion:
+    """Round-trip conversion tests for Braket → Qiskit → Braket and OpenQASM → Qiskit → Braket."""
+
+    # Shared OpenQASM programs for parameterized round-trip tests
+    _SINGLE_BOX_QASM = """
+OPENQASM 3.0;
+#pragma braket verbatim
+box {
+    h $0;
+    cnot $0, $1;
```
nilsquet, line 53: Nit: we could add another test (or modify this one) to also use qubits that are not consecutive
```suggestion
    cnot $2, $4;
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +312,78 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                if isinstance(symbol_type.size, IntegerLiteral):
+                    size = symbol_type.size.value
+                else:
+                    # Size is an Identifier or expression, can't determine size yet
+                    # This happens for function parameters like bit[n] where n is a variable
+                    return
+            else:
+                size = 1
+            
+            self._circuit.add_bits([Clbit() for _ in range(size)])
```
nilsquet, line 382: Could you add a test that checks the number of classical bits to see that there is no unintended side-effect of this method with the `self._circuit.add_register(num_qubits, num_qubits)` call above?

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,470 @@
...
+}
+y $1;
+"""
+
+    @staticmethod
+    def _to_qiskit_input(source, use_program):
```
nilsquet: I would move the helper methods to the top.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -740,6 +876,128 @@ def _add_instructions_no_parameter_restrictions(
...
+        # Check if this is a barrier with the verbatim box name label
+        if isinstance(operation, Barrier) and getattr(operation, "label", None) == verbatim_box_name:
+            barrier_count += 1
+
+            # Get the next verbatim box
+            try:
+                box_circuit, _ = next(verbatim_box_iter)
+            except StopIteration:
+                raise ValueError(
+                    f"Found more barriers with label '{verbatim_box_name}' than verbatim boxes. "
```
yitchen-tim: This is a good docstring for developers, but confusing for end users

yitchen-tim: same comment for fewer barriers

qiskit_braket_provider/providers/adapter.py
```diff
@@ -295,19 +312,78 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+    
+    This context extends AbstractProgramContext to build Qiskit QuantumCircuits from
+    OpenQASM 3 programs. It supports Braket verbatim pragmas, which are converted to
+    Qiskit BoxOp operations to preserve gate sequences that should not be optimized.
+    
+    Verbatim boxes are represented using Qiskit's native BoxOp construct, which treats
+    a block of operations atomically. All verbatim boxes in a circuit use the same
+    configurable label name.
+    """
+    def __init__(self, verbatim_box_name: str = _BRAKET_VERBATIM_BOX_NAME):
```
yitchen-tim, line 325: Will we ever need to change the name ? It's thoughtful to add customization, but given the the class itself is a private class, I feel like a global naming would suffice and it's makes the class and its methods cleaner. For example, no code change is needed for `to_qiskit` that way.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Nice find, fixed this in the revision

qiskit_braket_provider/providers/adapter.py
```diff
@@ -831,6 +1099,44 @@ def to_braket(
...
+        raise ValueError("Cannot have a Barrier labeled with the same label used for verbatim boxes")
+
+    # Check if any circuits have verbatim boxes and extract them before transpilation
+    has_verbatim_boxes = any(
+        any(
+            isinstance(instr.operation, BoxOp)
+            and getattr(instr.operation, "label", None) == verbatim_box_name
+            for instr in circ.data
+        )
+        for circ in circuits
```
nilsquet: Can we combine those two loops to not iterate the circuits twice?

qiskit_braket_provider/providers/adapter.py
aniksd-braket: I do like having that customization, if the user or us on the service side want to change it.

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,486 @@
...
+#pragma braket verbatim
+box {
+    h $0;
+    cnot $0, $1;
+}
+y $1;
+"""
+
+
+# --- Extract verbatim boxes tests ---
```
laurencap: Delete banners. If they're *really* necessary, it means your file is too long and it should be split into multiple files.

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,486 @@
...
+    """Extract (name, target) list from a Braket circuit."""
+    return [(instr.operator.name, instr.target) for instr in braket_circuit.instructions]
+
+
+def _to_qiskit_input(source, use_program):
+    """Return either a Program object or raw string for to_qiskit."""
+    return Program(source=source) if use_program else source
+
+
+# Shared OpenQASM programs for round-trip tests
```
laurencap: Use fixtures for test constants

qiskit_braket_provider/providers/adapter.py
```diff
@@ -740,6 +879,120 @@ def _add_instructions_no_parameter_restrictions(
...
+                reconstructed_circuit.append(
+                    box_instruction.operation, qubit_indices, clbit_indices
+                )
+        else:
+            # Get indices of qubits and clbits and add instruction as-is
+            qubit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.qubits]
+            clbit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.clbits]
+            reconstructed_circuit.append(operation, qubit_indices, clbit_indices)
+
+    # Check if any verbatim boxes remain
```
laurencap: Comments should be used sparingly. If code is clear on its own, then it shouldn't need comments. Imo, needing a comment usually means the code isn't clear enough, but in most of these cases the code is clear enough on its own (which means this was probably an AI agent explaining its work to you, and should be deleted by you if you've already reviewed it)

The reason comments can be so bad is that they're not testable code. So, as the code changes, the comments are at great risk for becoming outdated. There's no way to make sure that they stay up-to-date, and we always prefer mechanisms (like tests) over best-effort (like remembering to update the comments). That saves us from this PEP8 warning: "Comments that contradict the code are worse than no comments." No comments + easy to read code => more maintainable code

https://peps.python.org/pep-0008/#comments

laurencap (changes requested):

tests/providers/test_adapter_to_braket_verbatim.py
```diff
@@ -0,0 +1,470 @@
...
+    @pytest.mark.parametrize(
+        "verbatim, layout_method",
+        [
+            (True, None),
+            (False, None),
+            (False, "dense"),
+        ],
+        ids=["verbatim_true", "trivial_layout", "layout_override"],
+    )
+    def test_to_braket_verbatim_and_layout_options(self, verbatim, layout_method):
```
yitchen-tim: This is not a strong test. It's not testing verbatim box is really wrapping it. Even without verbatim, circuit can still be compiled as H and CNOT.

yitchen-tim (dismissed):

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Understood! I have removed all comments that seem excessive.

nilsquet (dismissed):

qiskit_braket_provider/providers/adapter.py
```diff
@@ -330,16 +403,74 @@ def add_gate_instruction(
...
+
+        # Ensure circuit has enough qubits for the target indices
+        # This is needed when using physical qubits ($0, $1, etc.) where no qubit register is declared
+        max_qubits = (max(target) + 1) if target else -1
+        self._circuit.add_bits([Qubit() for _ in range(max_qubits - self._circuit.num_qubits)])
+        self.num_qubits = self._circuit.num_qubits        
+
+        if self._in_verbatim_box:
+            # Ensure verbatim circuit also has enough qubits
+            self._verbatim_circuit.add_bits([Qubit() for _ in range(max_qubits - self._verbatim_circuit.num_qubits)])
```
laurencap: Hard to follow logic. Since I think this was written by AI, let's update this to be more readable by devs. Let's maintain high standards -- even and especially -- for AI code.

This logic is repeated 3 times at least, can we also move it into a helper method?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -831,6 +1082,37 @@ def to_braket(
...
+                # Check if any circuits have verbatim boxes and extract them before transpilation
+                elif isinstance(instr.operation, BoxOp):
+                    has_verbatim_boxes = True
+    
+    if has_barriers_named_verbatim:
+        raise ValueError("Cannot have a Barrier labeled with the same label used for verbatim boxes")
+    
+    if pass_manager and has_verbatim_boxes:
+        raise ValueError(
+            "Custom pass_manager is not supported with verbatim boxes. Verbatim boxes require controlled transpilation to preserve gate ordering."
```
laurencap: Nit: line length. Also applies to line 1136. Is there a linter for this repo? Sometimes linters won't split long strings, but usually they do warn/error on them.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1194,22 +1493,54 @@ def _translate_pauli(pauli: Pauli, coeff: float = 1.0) -> BraketObservable:
...
+    
+    Examples:
+        Convert an OpenQASM 3 program with a verbatim box:
+        
+        >>> openqasm_program = '''
+        ... OPENQASM 3.0;
+        ... qubit[2] q;
+        ... #pragma braket verbatim
+        ... box {
+        ...     h q[0];
```
laurencap: Re-upping the comment Nils left. These should be physical qubits

laurencap (changes requested):

qiskit_braket_provider/providers/adapter.py
aniksd-braket: This was actually written by me haha. It's a single line, used slightly differently in each case - so using a helper function here might actually be overly convoluted for a single LOC. What I could do instead, is break this one liner to use some helper variables that make things more readable

laurencap (approved):

PR #288, opened 2026-03-31
Refactor: extract qiskit_compile() from to_braket()

aniksd-braket:
Extract compilation logic into a new private `_compile()` function that returns compiled Qiskit circuits (along with relevant metadata) without translating to Braket.

### Summary

- Add `_compile()` internal function that returns a `_CompilationContext` `dataclass `with compiled circuits and resolved compilation state
- Simplify to_braket() to call `_compile()` then translate

### DIFF
--- a/qiskit_braket_provider/providers/__init__.py
+++ b/qiskit_braket_provider/providers/__init__.py
@@ -24,6 +24,7 @@ Provider classes and functions
     to_qiskit
 """
 
+from .adapter import qiskit_compile as qiskit_compile
 from .adapter import to_braket as to_braket
 from .adapter import to_qiskit as to_qiskit
 from .braket_backend import AWSBraketBackend as AWSBraketBackend
--- a/qiskit_braket_provider/providers/__init__.py
+++ b/qiskit_braket_provider/providers/__init__.py
@@ -24,7 +24,6 @@ Provider classes and functions
     to_qiskit
 """
 
-from .adapter import qiskit_compile as qiskit_compile
 from .adapter import to_braket as to_braket
 from .adapter import to_qiskit as to_qiskit
 from .braket_backend import AWSBraketBackend as AWSBraketBackend
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -11,7 +11,8 @@ from collections import defaultdict
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from math import inf, pi, prod
 from numbers import Number
-from typing import NamedTuple, TypeVar
+from dataclasses import dataclass
+from typing import TypeVar
 
 import numpy as np
 import qiskit.circuit.library as qiskit_gates
@@ -982,8 +983,9 @@ def _restore_verbatim_boxes(
     return reconstructed_circuit
 
 
-class _QiskitCompilationContext(NamedTuple):
-    """Internal result from _qiskit_compile containing compiled circuits and resolved state."""
+@dataclass(frozen=True)
+class _CompilationContext:
+    """Internal result from _compile containing compiled circuits and resolved state."""
 
     circuits: list[QuantumCircuit]
     single_instance: bool
@@ -995,7 +997,7 @@ class _QiskitCompilationContext(NamedTuple):
     pass_manager: PassManager | None
 
 
-def _qiskit_compile(
+def _compile(
     circuits: _Translatable | Iterable[_Translatable] = None,
     *args,
     qubit_labels: Sequence[int] | None = None,
@@ -1016,35 +1018,8 @@ def _qiskit_compile(
     layout_method: str | None = None,
     routing_method: str | None = None,
     seed_transpiler: int | None = None,
-) -> _QiskitCompilationContext:
-    """
-        Compiles Qiskit circuits by transpiling and processing verbatim boxes.
+) -> _CompilationContext:
 
-    Args:
-        circuits (QuantumCircuit | Circuit | Program | str | Iterable): Qiskit or Braket
-            circuit(s) or OpenQASM 3 program(s) to compile.
-        qubit_labels (Sequence[int] | None): Qubit indices on the Braket device. Default: ``None``.
-        target (Target | None): A backend transpiler target. Default: ``None``.
-        verbatim (bool): Whether to skip transpilation. Default: ``False``.
-        basis_gates (Sequence[str] | None): Gateset to transpile to. Default: ``None``.
-        coupling_map (list[list[int]] | None): Coupling map for transpilation. Default: ``None``.
-        angle_restrictions: Gate parameter angle constraints. Default: ``None``.
-        optimization_level (int): Qiskit transpile optimization level (0-3). Default: 0.
-        callback (Callable | None): Transpiler pass callback. Default: ``None``.
-        num_processes (int | None): Max parallel transpilation processes. Default: ``None``.
-        pass_manager (PassManager): Custom pass manager. Default: ``None``.
-        braket_device (Device): Braket device to transpile to. Default: ``None``.
-        add_measurements (bool): Whether to add measurements. Default: True.
-        circuit: DEPRECATED: use ``circuits``. Default: ``None``.
-        connectivity (list[list[int]] | None): DEPRECATED: use ``coupling_map``. Default: ``None``.
-        verbatim_box_name (str): Label for verbatim BoxOp operations. Default: ``"verbatim"``.
-        layout_method (str | None): Layout method for transpilation. Default: ``None``.
-        routing_method (str | None): Routing method for transpilation. Default: ``None``.
-        seed_transpiler (int | None): Seed for the transpiler. Default: ``None``.
-
-    Returns:
-        _QiskitCompilationContext
-    """
     circuits, single_instance = _get_circuits(circuits, circuit, add_measurements)
     if len(args) > 4:
         raise ValueError(f"Unknown arguments passed: {args[4:]}")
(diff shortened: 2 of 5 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -982,7 +982,20 @@ def _restore_verbatim_boxes(
     return reconstructed_circuit
 
 
-def to_braket(
+class _CompileResult(NamedTuple):
```
nilsquet: Since this class holds more than just the result, I'd prefer to call it something like `QiskitCompilationContext` or similar.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1176,13 +1120,187 @@ def to_braket(
...
+    pass_manager: PassManager | None = None,
+    braket_device: Device | None = None,
+    add_measurements: bool = True,
+    circuit: _Translatable | Iterable[_Translatable] | None = None,
+    connectivity: list[list[int]] | None = None,
+    verbatim_box_name: str = _BRAKET_VERBATIM_BOX_NAME,
+    layout_method: str | None = None,
+    routing_method: str | None = None,
+) -> QuantumCircuit | list[QuantumCircuit]:
+    """Compiles Qiskit circuits by transpiling and processing verbatim boxes.
```
nilsquet: I think we should add more information to this docstring similar to the level of detail we have for to_braket.

qiskit_braket_provider/providers/__init__.py
```diff
@@ -24,6 +24,7 @@
     to_qiskit
 """
```
nilsquet, line 25: We should list `qiskit_compile` as well.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1176,13 +1120,187 @@ def to_braket(
...
+def qiskit_compile(
+    circuits: _Translatable | Iterable[_Translatable] = None,
+    *args,
+    qubit_labels: Sequence[int] | None = None,
+    target: Target | None = None,
+    verbatim: bool | None = None,
+    basis_gates: Sequence[str] | None = None,
+    coupling_map: list[list[int]] | None = None,
+    angle_restrictions: Mapping[str, Mapping[int, set[float] | tuple[float, float]]] | None = None,
+    optimization_level: int = 0,
```
nilsquet: Could we add an optional `seed` for the `transpile` call?

tests/providers/test_adapter.py
```diff
@@ -1319,6 +1319,84 @@ def test_translate_sparse_pauli_op(self):
         )
 
 
+
+
+class TestQiskitCompile(TestCase):
```
nilsquet: I don't see a benefit of having this test class over the single tests and suggest to remove it.

tests/providers/test_adapter.py
```diff
@@ -1319,6 +1319,84 @@ def test_translate_sparse_pauli_op(self):
...
+    """Tests for the qiskit_compile function."""
+
+    def test_single_circuit(self):
+        """Tests qiskit_compile returns a single QuantumCircuit for single input."""
+        circuit = QuantumCircuit(2)
+        circuit.h(0)
+        circuit.cx(0, 1)
+        result = qiskit_compile(circuit)
+        self.assertIsInstance(result, QuantumCircuit)
```
nilsquet: Please add some more specific assertions for all tests.

tests/providers/test_adapter.py
```diff
@@ -1319,6 +1319,84 @@ def test_translate_sparse_pauli_op(self):
...
+        gate_names = [instr.operation.name for instr in result.data]
+        self.assertEqual(gate_names, ["h"])
+
+    def test_consistent_with_to_braket(self):
+        """Tests qiskit_compile output translates to same Braket circuit as to_braket."""
+        circuit = QuantumCircuit(2)
+        circuit.h(0)
+        circuit.cx(0, 1)
+
+        braket_direct = to_braket(circuit)
```
nilsquet: Could we use some more parameters for both calls to have a broader coverage?

nilsquet (changes requested):
Thanks @aniksd-braket! Having `qiskit_compile` as a public API will be really useful.

I think we can simplify the structure a bit. Right now `qiskit_compile` and `to_braket` are both thin wrappers around `_qiskit_compile`, with a `_CompileResult` NamedTuple to shuttle resolved state back for translation. This means the same 18-parameter signature appears three times.

We could make `qiskit_compile` the real compilation function (move the body of `_qiskit_compile` into it) and have it return `_CompileResult` directly. Then `to_braket` calls `qiskit_compile` and uses the result for translation. That eliminates `_qiskit_compile` and brings the signature count down from three to two. The tradeoff is that external callers who just want circuits would need to access `.circuits` on the result, but I think that's worth it for a cleaner internal structure.

Also, the docstrings for the same parameters differ between `qiskit_compile` and `to_braket` — for example `to_braket` documents the `optimization_level` breakdown (0-3) and `Raises` section, while `qiskit_compile` has abbreviated one-liners. Since they share the same parameters, the documentation should be consistent.

And longer term, with `qiskit_compile` owning compilation, `to_braket` could eventually become a pure translation function — compiled Qiskit circuit in, Braket circuit out. The canonical user flow would be `qiskit_compile` → `to_braket`. That's a breaking change to how `to_braket` works today, so we'd get there gradually by deprecating the compilation-related parameters on `to_braket` and pointing users to `qiskit_compile`.

What do you think?

aniksd-braket:
> Thanks @aniksd-braket! Having `qiskit_compile` as a public API will be really useful.
> 
> I think we can simplify the structure a bit. Right now `qiskit_compile` and `to_braket` are both thin wrappers around `_qiskit_compile`, with a `_CompileResult` NamedTuple to shuttle resolved state back for translation. This means the same 18-parameter signature appears three times.
> 
> We could make `qiskit_compile` the real compilation function (move the body of `_qiskit_compile` into it) and have it return `_CompileResult` directly. Then `to_braket` calls `qiskit_compile` and uses the result for translation. That eliminates `_qiskit_compile` and brings the signature count down from three to two. The tradeoff is that external callers who just want circuits would need to access `.circuits` on the result, but I think that's worth it for a cleaner internal structure.
> 
> Also, the docstrings for the same parameters differ between `qiskit_compile` and `to_braket` — for example `to_braket` documents the `optimization_level` breakdown (0-3) and `Raises` section, while `qiskit_compile` has abbreviated one-liners. Since they share the same parameters, the documentation should be consistent.
> 
> And longer term, with `qiskit_compile` owning compilation, `to_braket` could eventually become a pure translation function — compiled Qiskit circuit in, Braket circuit out. The canonical user flow would be `qiskit_compile` → `to_braket`. That's a breaking change to how `to_braket` works today, so we'd get there gradually by deprecating the compilation-related parameters on `to_braket` and pointing users to `qiskit_compile`.
> 
> What do you think?

Okay yes I am cool with that. That was my original design, but I moved to this one to keep the `qiskit_compile()` api clean and only return the compiled qiskit circuits.

sesmart:
I agree with @nilsquet, but I don't actually understand why you would need to add an additional api at all - right now you can just use the converted Backend to get a target, and then compile against that. So the understood trajectory would be to_qiskit -> transpile -> compiled circuit? This might just be a bit more confusing to add to an already bulky file. 

I also agree probably separating compilation from conversion makes sense, and a refactor definitely makes sense at some point, but the compilation is supposed to be pretty lightweight here - and I hope would be reproducable also using to_qiskit -> transpile -> to_braket. Can we just do it with a flag perhaps? Or an alternate return from the current functions?

qiskit_braket_provider/providers/__init__.py
speller26: I'd argue that `_qiskit_compile` should be private, and hidden from the autodoc

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Now that is is private, I think we do not need an excessive doc string

tests/providers/test_adapter.py
aniksd-braket: Removed the function so these tests are not required anymore

tests/providers/test_adapter.py
aniksd-braket: These tests are not required anymore

nilsquet (commented):
Thanks a lot for the update, @aniksd-braket. Just three minor things:
(1) the docstrings should match for both methods
(2) the PR description must be updated
(3) we could drop the "Qiskit" both in the method and in the `CompilationContext` because with the verbatim handling, we do significantly more than just calling Qiskit.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1002,76 +1015,35 @@ def to_braket(
...
-    """Converts a single or list of Qiskit QuantumCircuits to a single or list of Braket Circuits.
-
-    The recommended way to use this method is to minimally pass in qubit labels and a target
-    (instead of basis gates and coupling map). This ensures that the translated circuit is actually
-    supported by the device (and doesn't, for example, include unsupported parameters for gates).
-    The latter guarantees that the output Braket circuit uses the qubit labels of the Braket device,
-    which are not necessarily contiguous.
+    seed_transpiler: int | None = None,
+) -> _QiskitCompilationContext:
+    """
```
rmshaffer: We technically do not require docstrings for internal (`_`-prefixed) methods. In this case, since the docstring mostly duplicates the `to_braket` docstring, it might be cleaner to remove it.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -982,7 +982,20 @@ def _restore_verbatim_boxes(
     return reconstructed_circuit
 
 
-def to_braket(
+class _QiskitCompilationContext(NamedTuple):
```
speller26: Why are you subclassing `NamedTuple`?

speller26: And why not make this a `dataclass`?

aniksd-braket: Yup can just do that

nilsquet (approved):

PR #290, opened 2026-04-07
refactor verbatim pass handling

nilsquet:
## Refactor: verbatim box handling as transpiler passes

Replace _extract_verbatim_boxes / _restore_verbatim_boxes with proper Qiskit TransformationPass subclasses that operate on the DAG and compose into a StagedPassManager.

### Changes

- **verbatim_passes.py** (new): ExtractVerbatimBoxes and RestoreVerbatimBoxes passes
- **adapter.py**: Delete old functions, replace transpile() with generate_preset_pass_manager() with verbatim passes injected  at pre_init / post_optimization
- **Tests**: Updated to use new passes; added 10 dedicated pass tests

### Why

The old functions manually rebuilt circuits instruction-by-instruction outside the transpiler pipeline. The new passes use dag.substitute_node / dag.substitute_node_with_dag, compose naturally into any PassManager, and are reusable independently of to_braket().

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -9,7 +9,6 @@ sequences that should not be optimized.
 import warnings
 from collections import defaultdict
 from collections.abc import Callable, Iterable, Mapping, Sequence
-from dataclasses import dataclass
 from math import inf, pi, prod
 from numbers import Number
 from typing import TypeVar
@@ -17,7 +16,7 @@ from typing import TypeVar
 import numpy as np
 import qiskit.circuit.library as qiskit_gates
 import qiskit.quantum_info as qiskit_qi
-from qiskit import QuantumCircuit, transpile
+from qiskit import QuantumCircuit
 from qiskit.circuit import (
     Barrier,
     BoxOp,
@@ -43,6 +42,7 @@ from qiskit.transpiler import (
     Target,
     TransformationPass,
 )
+from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
 from qiskit_ionq import add_equivalences, ionq_gates
 from sympy import Add, Expr, Mul, Pow, Symbol
 
@@ -101,6 +101,10 @@ from braket.ir.openqasm.modifiers import Control
 from braket.parametric import FreeParameter, FreeParameterExpression, Parameterizable
 from qiskit_braket_provider.exception import QiskitBraketException
 from qiskit_braket_provider.providers import braket_instructions
+from qiskit_braket_provider.providers.verbatim_passes import (
+    ExtractVerbatimBoxes,
+    RestoreVerbatimBoxes,
+)
 
 add_equivalences()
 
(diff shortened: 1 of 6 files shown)


qiskit_braket_provider/providers/verbatim_passes.py
```diff
@@ -0,0 +1,130 @@
...
+        self._verbatim_box_name = verbatim_box_name
+
+    def run(self, dag: DAGCircuit) -> DAGCircuit:
+        """Replace matching ``BoxOp`` nodes with labeled barriers.
+
+        Raises:
+            ValueError: If the DAG already contains a barrier whose label
+                matches ``verbatim_box_name``.
+        """
+        for node in dag.op_nodes(Barrier):
```
sesmart, line 64: I think topological_op_nodes() [is recommended](https://quantum.cloud.ibm.com/docs/en/api/qiskit/transpiler#randomness-and-determinism) for determinism.

qiskit_braket_provider/providers/verbatim_passes.py
```diff
@@ -0,0 +1,130 @@
...
+            if _is_verbatim_label(getattr(node.op, "label", None), self._verbatim_box_name):
+                raise ValueError(
+                    f"Circuit contains a Barrier with label '{node.op.label}' "
+                    "which conflicts with the verbatim box label"
+                )
+
+        verbatim_boxes = {}
+        index = 0
+        for node in dag.op_nodes(BoxOp):
```
sesmart, line 73: Similar to above.

qiskit_braket_provider/providers/verbatim_passes.py
sesmart: Though for the validation here it doesn't really matter.

qiskit_braket_provider/providers/verbatim_passes.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Raises:
+            ValueError: If the number of labeled barriers does not match
+                the number of stashed verbatim boxes.
+        """
+        verbatim_boxes = self.property_set.get("verbatim_boxes", {})
+        if not verbatim_boxes:
+            return dag
+
+        for node in dag.op_nodes(Barrier):
```
sesmart, line 110: Replace with `topological_op_nodes()`

qiskit_braket_provider/providers/verbatim_passes.py
```diff
@@ -0,0 +1,130 @@
...
+            label = getattr(node.op, "label", None)
+            if not _is_verbatim_label(label, self._verbatim_box_name):
+                continue
+            if label not in verbatim_boxes:
```
sesmart, line 114: Would an AnalysisPass for the barrier check make sense? I don't think it is necessary, but curious what you think.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1324,171 +1179,43 @@ def _compile(
...
-                routing_method=effective_routing_method,
+                layout_method="trivial" if has_verbatim_boxes else layout_method,
+                routing_method="none" if has_verbatim_boxes else routing_method,
                 seed_transpiler=seed_transpiler,
             )
+            if has_verbatim_boxes:
+                pm.pre_init = PassManager([ExtractVerbatimBoxes(verbatim_box_name)])
+                pm.post_optimization = PassManager([RestoreVerbatimBoxes(verbatim_box_name)])
+            circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
+    elif has_verbatim_boxes:
```
sesmart, line 1204: I really really like this approach generally. Wondering if some of the conditional logic could use the [ConditionalController](https://github.com/Qiskit/qiskit/blob/51680f9d869b189b964e9a523e2292221fd65287/qiskit/transpiler/preset_passmanagers/common.py#L345), or if makes it more complex in this case?

nilsquet:
implemented as part of #356

PR #294, opened 2026-04-10
feat: Changes to support branching statements using midcircuit measurements

aniksd-braket:
### What
Converts OpenQASM 3 if/else statements conditioned on measurement results into Qiskit IfElseOp operations during the OpenQASM → Qiskit conversion.

### Changes

qiskit_braket_provider/providers/adapter.py
- `supports_midcircuit_measurement` → returns True, enabling the interpreter to delegate control flow to the context
- `set_visitor()` — stores the interpreter's visit method for use in control flow handlers
- `handle_branching_statement()` — tries static evaluation first; if the condition can't be resolved (measurement result), creates separate QuantumCircuit 
objects for each branch by swapping self._circuit, visits the blocks via the stored visitor, then wraps them in an IfElseOp
- `handle_for_loop()` / `handle_while_loop()` — replicate the interpreter's default loop logic so `supports_midcircuit_measurement=True` doesn't break non-MCM loops,
while still allowing MCM branches inside loop bodies
- `_resolve_condition()` / `_resolve_clbit_index()` — map OpenQASM AST condition nodes to Qiskit (Clbit, int) tuples using a `_clbit_offset` map populated during 
`declare_variable()`
- `declare_variable()` — now tracks the starting clbit index for each bit variable in `_clbit_offset`

### Design decisions
- Static vs MCM detection uses try/except on condition evaluation rather than variable tracking — avoids false positives from input bit variables
- Branch circuits are built by temporarily swapping `self._circuit`, so all existing instruction routing (`add_gate_instruction`, `add_measure`, etc.) works without 
modification
- Loop handlers replicate interpreter logic directly (rather than toggling `supports_midcircuit_measurement`) so MCM branches inside loops are handled correctly

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -9,6 +9,7 @@ sequences that should not be optimized.
 import warnings
 from collections import defaultdict
 from collections.abc import Callable, Iterable, Mapping, Sequence
+from copy import deepcopy
 from dataclasses import dataclass
 from math import inf, pi, prod
 from numbers import Number
@@ -25,6 +26,7 @@ from qiskit.circuit import (
     Clbit,
     ControlledGate,
     Gate,
+    IfElseOp,
     Measure,
     Parameter,
     ParameterExpression,
@@ -52,13 +54,28 @@ from braket.circuits import Observable as BraketObservable
 from braket.circuits import gates as braket_gates
 from braket.circuits import noises as braket_noises
 from braket.circuits import observables as braket_observables
+from braket.default_simulator.openqasm._helpers.arrays import convert_range_def_to_range
+from braket.default_simulator.openqasm._helpers.casting import cast_to
 from braket.default_simulator.openqasm.interpreter import Interpreter, VerbatimBoxDelimiter
 from braket.default_simulator.openqasm.parser.openqasm_ast import (
+    BinaryExpression,
     BitType,
+    BooleanLiteral,
+    BranchingStatement,
     ClassicalType,
+    ForInLoop,
+    Identifier,
+    IndexedIdentifier,
+    IndexExpression,
     IntegerLiteral,
+    RangeDefinition,
+    WhileLoop,
 )
-from braket.default_simulator.openqasm.program_context import AbstractProgramContext
+from braket.default_simulator.openqasm.program_context import (
+    AbstractProgramContext,
+    _BreakSignal,
+    _ContinueSignal,
+)
 from braket.device_schema import (
     DeviceActionType,
     DeviceCapabilities,
@@ -336,6 +353,8 @@ class _QiskitProgramContext(AbstractProgramContext):
         self._in_verbatim_box = False
         self._verbatim_circuit: QuantumCircuit | None = None
         self._verbatim_box_name = verbatim_box_name
+        self._visitor: Callable | None = None
+        self._clbit_offset: dict[str, int] = {}
 
     @property
     def circuit(self):
@@ -380,6 +399,8 @@ class _QiskitProgramContext(AbstractProgramContext):
             else:
                 size = 1
             
+            # this is used deal with Qiskit circuit storing all classical bits in a flat list
+            self._clbit_offset[name] = self._circuit.num_clbits
             self._circuit.add_bits([Clbit() for _ in range(size)])
 
     def is_builtin_gate(self, name: str) -> bool:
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+    def handle_while_loop(self, node: WhileLoop) -> None:
+        while cast_to(BooleanLiteral, self._visitor(node.while_condition)).value:
+            try:
+                self._visitor(deepcopy(node.block))
+            except _BreakSignal:
+                break
+            except _ContinueSignal:
+                continue
+
+    def _resolve_condition(
```
nilsquet, line 571: `_resolve_condition` has two issues:

1. **Bare conditions crash** — `if (c[0])` is valid OpenQASM 3 but throws `AttributeError` because the condition is an `IndexExpression`, not a `BinaryExpression`, so `.lhs` doesn't exist.

2. **`!=` silently produces wrong results** — `if (c[0] != 0)` becomes `if (c[0] == 0)` because the operator is never checked.

Suggested fix:

```python
def _resolve_condition(self, condition) -> tuple[Clbit, int]:
    # Bare identifier — if (c) or if (c[0]) means == 1
    if isinstance(condition, (Identifier, IndexedIdentifier, IndexExpression)):
        clbit_index = self._resolve_clbit_index(condition)
        return (self._circuit.clbits[clbit_index], 1)

    if condition.op.name != "==":
        raise NotImplementedError(f"Unsupported condition operator: {condition.op.name}")

    if isinstance(condition.lhs, (Identifier, IndexedIdentifier, IndexExpression)):
        clbit_index = self._resolve_clbit_index(condition.lhs)
        value = condition.rhs.value
    else:
        clbit_index = self._resolve_clbit_index(condition.rhs)
        value = condition.lhs.value
    return (self._circuit.clbits[clbit_index], int(value))
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        self, condition: BinaryExpression
+    ) -> tuple[Clbit, int]:
+        """Convert an OpenQASM condition AST node to a Qiskit (Clbit, int) condition."""
+        if isinstance(condition.lhs, (Identifier, IndexedIdentifier, IndexExpression)):
```
nilsquet, line 575:
```suggestion
        if isinstance(condition.lhs, (Identifier, IndexExpression)):
```

Nit: `IndexedIdentifier` in the isinstance checks in `_resolve_condition` is dead code. `IndexedIdentifier` is an AST node for lvalues — assignment targets (`c[0] = measure ...`) and gate targets (`h q[1]`). The parser never produces it in condition expressions. Conditions use `IndexExpression` for `c[0]`, `Identifier` for `c`, or `BinaryExpression` for `c[0] == 1`. Consider removing `IndexedIdentifier` from the checks to avoid confusion.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        return (self._circuit.clbits[clbit_index], int(value))
+
+    def _resolve_clbit_index(self, node: Identifier | IndexExpression) -> int:
+        """Resolve an identifier or indexed identifier to a classical bit index."""
+        if isinstance(node, IndexExpression):
+            name = node.collection.name
+            index = node.index[0].value
+        elif isinstance(node, Identifier):
+            name = node.name
+            index = 0
```
nilsquet, line 590: `if (c == 3)` where `c` is multi-bit (e.g. `bit[3] c`) silently becomes `if (c[0] == 3)` — always false since a single bit can only be 0 or 1. `_resolve_clbit_index` treats bare `Identifier` as index 0.

For now, we could check and throw an error:
```
    if self._clbit_size[name] > 1:
        raise NotImplementedError(f"Multi-bit register condition on '{name}' is not supported")
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -52,13 +54,28 @@
 from braket.circuits import gates as braket_gates
 from braket.circuits import noises as braket_noises
 from braket.circuits import observables as braket_observables
+from braket.default_simulator.openqasm._helpers.arrays import convert_range_def_to_range
+from braket.default_simulator.openqasm._helpers.casting import cast_to
 from braket.default_simulator.openqasm.interpreter import Interpreter, VerbatimBoxDelimiter
 from braket.default_simulator.openqasm.parser.openqasm_ast import (
+    BinaryExpression,
```
nilsquet, line 61: All the imports require `amazon-braket-default-simulator>=1.36.0`. Either we need to adjust the lower bound in the requirements or check here and throw an error if the installed version is too low.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+    def handle_branching_statement(self, node: BranchingStatement) -> None:
+        # Try static evaluation first; fall back to MCM if the condition
+        # can't be resolved (e.g. it depends on a measurement result) due to
+        # NameError (uninitialized variable, i.e. measurement result) or
+        # TypeError (can't cast to boolean)
+        try:
+            condition = cast_to(BooleanLiteral, self._visitor(node.condition))
+        except (NameError, TypeError):
+            pass
+        else:
```
nilsquet, line 515: Using try/except as control flow here makes the logic hard to follow. Consider checking `_clbit_offset` directly to decide the path — the intent becomes explicit:

```python
def _is_mcm_condition(self, condition) -> bool:
    if isinstance(condition, (Identifier, IndexExpression)):
        name = condition.name if isinstance(condition, Identifier) else condition.collection.name
        return name in self._clbit_offset
    if isinstance(condition, BinaryExpression):
        return self._is_mcm_condition(condition.lhs) or self._is_mcm_condition(condition.rhs)
    return False
```

Then:

```python
if self._is_mcm_condition(node.condition):
    # MCM path
else:
    # static path
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
             raise ValueError("Verbatim box created using invalid marker")
 
+    @property
+    def supports_midcircuit_measurement(self) -> bool:
+        return True
+
+    def set_visitor(self, visitor: Callable) -> None:
+        self._visitor = visitor
+
+    def handle_branching_statement(self, node: BranchingStatement) -> None:
```
nilsquet, line 506: Missing docstring.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+                self._visitor(statement)
+
+        self._circuit = main_circuit
+
+        if_else_op = IfElseOp(condition, true_body, false_body)
+        qubits = list(range(main_circuit.num_qubits))
+        clbits = list(range(main_circuit.num_clbits))
+        self._circuit.append(if_else_op, qubits, clbits)
+
+    def handle_for_loop(self, node: ForInLoop) -> None:
```
nilsquet, line 546: Missing docstring.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+            with self.enter_scope():
+                self.declare_variable(node.identifier.name, node.type, i)
+                try:
+                    self._visitor(deepcopy(node.block))
+                except _BreakSignal:
+                    break
+                except _ContinueSignal:
+                    continue
+
+    def handle_while_loop(self, node: WhileLoop) -> None:
```
nilsquet, line 562: Missing docstring.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+            if condition.value:
+                self._visitor(node.if_block)
+            elif node.else_block:
+                self._visitor(node.else_block)
+            return
+
+        condition = self._resolve_condition(node.condition)
+        main_circuit = self._circuit
+
+        # Visit if block into a separate circuit
```
nilsquet, line 525: This comment doesn't add value.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        true_body = QuantumCircuit(main_circuit.num_qubits, main_circuit.num_clbits)
+        self._circuit = true_body
+        for statement in node.if_block:
+            self._visitor(statement)
+
+        # Visit else block if present
```
nilsquet, line 531: This comment doesn't add value.

tests/providers/test_adapter_branching.py
```diff
@@ -0,0 +1,610 @@
...
+c[0] = measure q[0];
+"""
+    qc = to_qiskit(qasm)
+    h_count = sum(1 for name, _ in _get_ops_with_qubits(qc) if name == "h")
+    assert h_count == 0
+
+
+def test_resolve_clbit_index_unsupported_type():
+    from braket.default_simulator.openqasm.parser.openqasm_ast import IntegerLiteral
+    from qiskit_braket_provider.providers.adapter import _QiskitProgramContext
```
nilsquet, line 606: Should be moved to the top.

tests/providers/test_adapter_branching.py
```diff
@@ -0,0 +1,610 @@
+"""Tests for branching statement (if/else) support in the Qiskit adapter."""
```
nilsquet, line 1: I really like the tests, thanks for that! We should just add some more tests we expect to fail, e.g.,

- `if (c[0])` — bare condition without `==`
- `if (c[0] != 0)` — non-`==` operator
- `if (c == 3)` where `c` is multi-bit — full register condition
- Nested if/else — `if (c[0] == 1) { if (c[1] == 1) { ... } }`

tests/providers/test_adapter_branching.py
```diff
@@ -0,0 +1,610 @@
...
+from qiskit_braket_provider import to_qiskit
+from qiskit_braket_provider.providers.adapter import _compile
+
+
+def _get_if_else_ops(circuit):
+    """Extract all IfElseOp instructions from a circuit."""
+    return [instr for instr in circuit.data if isinstance(instr.operation, IfElseOp)]
+
+
+def _get_gate_names(circuit):
```
nilsquet, line 17: This function is never used.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -475,6 +496,103 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+                    self._visitor(deepcopy(node.block))
+                except _BreakSignal:
+                    break
+                except _ContinueSignal:
+                    continue
+
+    def handle_while_loop(self, node: WhileLoop) -> None:
+        while cast_to(BooleanLiteral, self._visitor(node.while_condition)).value:
+            try:
+                self._visitor(deepcopy(node.block))
```
nilsquet, line 565: `handle_for_loop` and `handle_while_loop` unroll loops into repeated gates instead of creating Qiskit's `ForLoopOp`/`WhileLoopOp`. For example, `for i in [0:2] { h q[0]; }` produces 3 separate `h` gates — the loop structure is lost.

This was already the behavior before the PR (the base interpreter unrolls too), but since `supports_midcircuit_measurement = True` delegates loop handling to these new methods, this is an opportunity to create proper loop ops instead. Qiskit supports both `ForLoopOp` and `WhileLoopOp`, same pattern as `IfElseOp`.

This also means `while` loops conditioned on measurement results crash:

```python
to_qiskit("""
OPENQASM 3.0;
qubit[2] q;
bit[1] c;
c[0] = measure q[0];
while (c[0] == 1) {
    h q[1];
    c[0] = measure q[0];
}
""")
# NameError: Identifier 'c' is not initialized.
```

A `WhileLoopOp` would handle this the same way `IfElseOp` handles `if`.

nilsquet (changes requested):

aniksd-braket:
Closing in lieu of #295 incorporating new interface

PR #295, opened 2026-04-15
Feature: handle branching using Qiskit

aniksd-braket:
### Summary

Replaces the old visitor-based `handle_branching_statement/handle_for_loop/handle_while_loop` methods with the new generator-based `evaluate_condition/
evaluate_for_range/evaluate_while_condition` interface from `AbstractProgramContext` in `amazon-braket-default-simulator-python`.

### What changed

`_QiskitProgramContext` in `adapter.py`:
- Uses a circuit stack (`_circuit_stack`) instead of mutating `self._circuit`. The root circuit is `_circuit_stack[0]`, and `_active_circuit` always points  to the top of the stack. All instruction-adding methods (`add_gate_instruction`, `add_measure`, `add_qubits`, etc.) operate on `_active_circuit`.
- `evaluate_condition` pushes sub-circuits onto the stack for `if/else` blocks, yields control back to the interpreter, then pops them off and constructs an `IfElseOp`.
- `evaluate_for_range` and `evaluate_while_condition` use a lightweight `_evaluate_expression` to evaluate loop ranges and conditions.
- Supports bare identifier conditions (if `(c)`, if `(c[0])`) treated as `== 1`.
- Validates that MCM conditions use `==` operator; raises `TypeError` for unsupported operators.
- Static conditions (e.g., if `(1 == 1)`) are evaluated directly without creating an `IfElseOp`.

Tests:
- New `test_adapter_branching.py` with 31 tests covering: `if/else` bodies, `if-only`, condition values, `clbit` resolution, circuit dimensions, gates  before/after branches, multiple branches, for/while loops with break/continue, static conditions, bare identifiers, unsupported operators, and compilation preservation.

Note:
- Loops are out of scope for this PR. The implementation here is just to ensure non-MCM loops are handled.

### Dependencies

- Companion PR in `amazon-braket-default-simulator-python`: [feature/pre-evaluate-branching-condition](https://github.com/amazon-braket/amazon-braket-
default-simulator-python/pull/new/feature/pre-evaluate-branching-condition) — adds interpreter pre-evaluation of complex conditions (e.g., 
FunctionCall, SizeOf) before delegating to the context. Without this change, `test_roundtrip_openqasm_program` is skipped.

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -353,8 +353,7 @@ class _QiskitProgramContext(AbstractProgramContext):
                 Default: "verbatim"
         """
         super().__init__()
-        self._circuit = QuantumCircuit()
-        self._circuit_stack: list[QuantumCircuit] = []
+        self._circuit_stack: list[QuantumCircuit] = [QuantumCircuit()]
         self._param_map = {}
         self._in_verbatim_box = False
         self._verbatim_circuit: QuantumCircuit | None = None
@@ -363,8 +362,8 @@ class _QiskitProgramContext(AbstractProgramContext):
 
     @property
     def _active_circuit(self) -> QuantumCircuit:
-        """The circuit that instructions should be added to (top of stack or main)."""
-        return self._circuit_stack[-1] if self._circuit_stack else self._circuit
+        """The circuit that instructions should be added to (top of stack)."""
+        return self._circuit_stack[-1]
 
     @property
     def circuit(self):
@@ -373,7 +372,7 @@ class _QiskitProgramContext(AbstractProgramContext):
                 "Unclosed verbatim box at end of program. "
                 "Every verbatim box start marker must have a matching end marker."
             )
-        return self._circuit
+        return self._circuit_stack[0]
 
     def add_qubits(self, name: str, num_qubits: int | None = 1) -> None:
         super().add_qubits(name, num_qubits)
@@ -525,7 +524,10 @@ class _QiskitProgramContext(AbstractProgramContext):
         # Try static evaluation first
         try:
             result = cast_to(BooleanLiteral, self._evaluate_expression(condition))
-        except (NameError, TypeError, ValueError, AttributeError):
+        except (TypeError, ValueError, AttributeError):
+            # TypeError: unsupported node type (e.g., FunctionCall) in _evaluate_expression
+            # ValueError: binary expression with None operand (unresolved measurement)
+            # AttributeError: cast_to on None.value (bare bit with pending measurement)
             pass
         else:
             yield result.value
@@ -556,7 +558,8 @@ class _QiskitProgramContext(AbstractProgramContext):
         yield True
         self._circuit_stack.pop()
 
-        # Push circuit for else-block
+        # Push circuit for else-block (the interpreter always consumes this yield
+        # even for if-only blocks; the empty circuit is discarded below)
         false_body = QuantumCircuit(main.num_qubits, main.num_clbits)
         self._circuit_stack.append(false_body)
         yield False
@@ -627,14 +630,14 @@ class _QiskitProgramContext(AbstractProgramContext):
         else:
             clbit_index = self._resolve_clbit_index(condition.rhs)
             value = condition.lhs.value
-        return (self._circuit.clbits[clbit_index], int(value))
+        return (self.circuit.clbits[clbit_index], int(value))
 
     def _resolve_condition_from_identifier(
         self, condition: Identifier | IndexExpression
     ) -> tuple[Clbit, int]:
         """Convert a bare identifier condition (e.g., `c` or `c[0]`) to (Clbit, 1)."""
         clbit_index = self._resolve_clbit_index(condition)
-        return (self._circuit.clbits[clbit_index], 1)
+        return (self.circuit.clbits[clbit_index], 1)
 
     def _resolve_clbit_index(self, node: Identifier | IndexExpression) -> int:
         """Resolve an identifier or indexed identifier to a classical bit index."""
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -9,6 +9,7 @@ sequences that should not be optimized.
 import warnings
 from collections import defaultdict
 from collections.abc import Callable, Iterable, Mapping, Sequence
+from copy import deepcopy
 from dataclasses import dataclass
 from math import inf, pi, prod
 from numbers import Number
@@ -25,6 +26,7 @@ from qiskit.circuit import (
     Clbit,
     ControlledGate,
     Gate,
+    IfElseOp,
     Measure,
     Parameter,
     ParameterExpression,
@@ -52,13 +54,28 @@ from braket.circuits import Observable as BraketObservable
 from braket.circuits import gates as braket_gates
 from braket.circuits import noises as braket_noises
 from braket.circuits import observables as braket_observables
+from braket.default_simulator.openqasm._helpers.arrays import convert_range_def_to_range
+from braket.default_simulator.openqasm._helpers.casting import cast_to
 from braket.default_simulator.openqasm.interpreter import Interpreter, VerbatimBoxDelimiter
 from braket.default_simulator.openqasm.parser.openqasm_ast import (
+    BinaryExpression,
     BitType,
+    BooleanLiteral,
+    BranchingStatement,
     ClassicalType,
+    ForInLoop,
+    Identifier,
+    IndexedIdentifier,
+    IndexExpression,
     IntegerLiteral,
+    RangeDefinition,
+    WhileLoop,
 )
-from braket.default_simulator.openqasm.program_context import AbstractProgramContext
+from braket.default_simulator.openqasm.program_context import (
+    AbstractProgramContext,
+    _BreakSignal,
+    _ContinueSignal,
+)
 from braket.device_schema import (
     DeviceActionType,
     DeviceCapabilities,
@@ -336,6 +353,8 @@ class _QiskitProgramContext(AbstractProgramContext):
         self._in_verbatim_box = False
         self._verbatim_circuit: QuantumCircuit | None = None
         self._verbatim_box_name = verbatim_box_name
+        self._visitor: Callable | None = None
+        self._clbit_offset: dict[str, int] = {}
 
     @property
     def circuit(self):
@@ -380,6 +399,8 @@ class _QiskitProgramContext(AbstractProgramContext):
             else:
                 size = 1
             
+            # this is used deal with Qiskit circuit storing all classical bits in a flat list
+            self._clbit_offset[name] = self._circuit.num_clbits
             self._circuit.add_bits([Clbit() for _ in range(size)])
 
     def is_builtin_gate(self, name: str) -> bool:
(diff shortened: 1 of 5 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+            case list():
+                return [self._evaluate_expression(item) for item in expression]
+            case _:
+                raise TypeError(f"Cannot evaluate expression of type {type(expression).__name__}")
+
+    def _resolve_condition(
+        self, condition: BinaryExpression
+    ) -> tuple[Clbit, int]:
+        """Convert an OpenQASM condition AST node to a Qiskit (Clbit, int) condition."""
+        if isinstance(condition.lhs, (Identifier, IndexedIdentifier, IndexExpression)):
```
nilsquet: see https://github.com/qiskit-community/qiskit-braket-provider/pull/294#discussion_r3073141731

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        return (self.circuit.clbits[clbit_index], 1)
+
+    def _resolve_clbit_index(self, node: Identifier | IndexExpression) -> int:
+        """Resolve an identifier or indexed identifier to a classical bit index."""
+        if isinstance(node, IndexExpression):
+            name = node.collection.name
+            index = node.index[0].value
+        elif isinstance(node, Identifier):
+            name = node.name
+            index = 0
```
nilsquet, line 677: see https://github.com/qiskit-community/qiskit-braket-provider/pull/294#discussion_r3073173262

qiskit_braket_provider/providers/adapter.py
```diff
@@ -52,13 +53,34 @@
...
 from braket.circuits import observables as braket_observables
+from braket.default_simulator.openqasm._helpers.arrays import convert_range_def_to_range
+from braket.default_simulator.openqasm._helpers.casting import cast_to
+from braket.default_simulator.openqasm._helpers.functions import (
+    evaluate_binary_expression,
+    evaluate_unary_expression,
+)
 from braket.default_simulator.openqasm.interpreter import Interpreter, VerbatimBoxDelimiter
 from braket.default_simulator.openqasm.parser.openqasm_ast import (
+    ArrayLiteral,
```
nilsquet, line 64: see https://github.com/qiskit-community/qiskit-braket-provider/pull/294#discussion_r3073181805

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        and yields only the taken branch.
+        """
+        # Try static evaluation first
+        try:
+            result = cast_to(BooleanLiteral, self._evaluate_expression(condition))
+        except (TypeError, ValueError, AttributeError):
+            # TypeError: unsupported node type (e.g., FunctionCall) in _evaluate_expression
+            # ValueError: binary expression with None operand (unresolved measurement)
+            # AttributeError: cast_to on None.value (bare bit with pending measurement)
+            pass
```
nilsquet: Same feedback as #294 (https://github.com/qiskit-community/qiskit-braket-provider/pull/294#discussion_r3073274801) — using try/except to distinguish static vs MCM conditions masks real bugs. `_clbit_offset already` has the info:                                                                                                                                                      
```                                                                                                                                                                                                                                                                                                                         
def _is_mcm_condition(self, condition) -> bool:                                                                                                                                                                                                                                                                 
    if isinstance(condition, Identifier):                                                                                                                                                                                                                                                                                    
        return condition.name in self._clbit_offset                                                                                                                                                                                                                                                                          
    if isinstance(condition, IndexExpression):                                                                                                                                                                                                                                                                               
        return condition.collection.name in self._clbit_offset                                                                                                                                                                                                                                                               
    if isinstance(condition, BinaryExpression):                                                                                                                                                                                                                                                                              
        return (self._references_measurement_result(condition.lhs)                                                                                                                                                                                                                                                           
                or self._references_measurement_result(condition.rhs))                                                                                                                                                                                                                                                       
    return False   
```

tests/providers/test_adapter_branching.py
```diff
@@ -0,0 +1,713 @@
...
+    FunctionCall,
+    IntegerLiteral,
+    UnaryExpression,
+    UnaryOperator,
+)
+from qiskit_braket_provider import to_qiskit
+from qiskit_braket_provider.providers.adapter import _compile, _QiskitProgramContext
+
+
+def _get_if_else_ops(circuit):
```
nilsquet, line 22: I suggest to add test for
 - if (c == 3) where c is multi-bit — full register condition
 - Nested if/else — if (c[0] == 1) { if (c[1] == 1) { ... } }

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        if isinstance(index, RangeDefinition):
+            index_values = [IntegerLiteral(x) for x in convert_range_def_to_range(index)]
+        else:
+            index_values = index.values
+        for i in index_values:
+            with self.enter_scope():
+                self.declare_variable(loop_var, loop_type, i)
+                yield
+
+    def evaluate_while_condition(self, condition):
```
nilsquet, line 608: Same feedback as #294 (https://github.com/qiskit-community/qiskit-braket-provider/pull/294#discussion_r3073397418) — loops unroll into repeated gates instead of creating  
  Qiskit ForLoopOp/WhileLoopOp.                                                                                                                                              
                                                                                                                                                                             
  MCM-conditioned while loops still crash — _evaluate_expression returns None for unresolved measurement results, then evaluate_binary_expression(None, ...) fails:          
                                                                                                                                                                             
  ```python                                                                                                                                                                  
  to_qiskit("""                                                                                                                                                              
  OPENQASM 3.0;                                                                                                                                                              
  qubit[2] q; bit[1] c;                                                                                                                                                      
  c[0] = measure q[0];                                                                                                                                                       
  while (c[0] == 1) { h q[1]; c[0] = measure q[0]; }                                                                                                                         
  """)                                                                                                                                                                       
```                                                                                                                                                                           

 If loop ops are out of scope, please document unrolling as a known limitation and add a test asserting the expected failure for MCM while loops.

File: qiskit_braket_provider/providers/adapter.py [hunk 7]
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+            return
+
+        # MCM path: resolve condition to (Clbit, value)
+        main = self._active_circuit
+        if isinstance(condition, (Identifier, IndexExpression)):
+            # Bare identifier like `if (c)` or `if (c[0])` — equivalent to `== 1`
+            resolved_condition = self._resolve_condition_from_identifier(condition)
+        elif isinstance(condition, BinaryExpression):
+            if condition.op != BinaryOperator["=="]:
+                raise TypeError(
```
nilsquet: Nit: 
```suggestion
                raise NotImplementedError(
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -405,32 +435,34 @@ def add_gate_instruction(
...
+        active = self._active_circuit
         # Ensure circuit has enough qubits for the target indices by adding missing qubits
         # This is needed when using physical qubits ($0, $1, etc.) where no qubit register is declared
         max_qubits = (max(target) + 1) if target else -1
-        num_missing_qubits = max_qubits - self._circuit.num_qubits
-        self._circuit.add_bits([Qubit() for _ in range(num_missing_qubits)])
-        self.num_qubits = self._circuit.num_qubits        
+        num_missing_qubits = max_qubits - active.num_qubits
+        active.add_bits([Qubit() for _ in range(num_missing_qubits)])
+        self.num_qubits = max(self.num_qubits, active.num_qubits)
```
nilsquet, line 445: If a gate inside a branch auto-creates qubits, the sub-circuit grows but main was captured before the branch, causing a confusing error:                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                   
  ```python                                                                                                                                                                                                                                                                                                                        
  to_qiskit("""OPENQASM 3.0;                                                                                                                                                                                                                                                                                                       
  qubit[2] q;                                                                                                                                                                                                                                                                                                                      
  bit[1] c;                                                                                                                                                                                                                                                                                                                        
  c[0] = measure q[0];                                                                                                                                                                                                                                                                                                             
  if (c[0] == 1) {                                                                                                                                                                                                                                                                                                                 
      h $4;                                                                                                                                                                                                                                                                                                                        
  }                                                                                                                                                                                                                                                                                                                                
  """)                                                                                                                                                                                                                                                                                                                             
# CircuitError: 'The amount of qubit arguments 2 does not match the instruction expectation (5).'                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                   The IfElseOp sub-circuit has 5 qubits but is wired with only 2 from main. Consider syncing main's qubit count to match the sub-circuits before appending the IfElseOp.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        Each yield pushes a new circuit onto the stack; after the interpreter
+        visits the block, the circuit is popped and used to build an IfElseOp.
+
+        For static conditions (no measurement dependency), evaluates directly
+        and yields only the taken branch.
+        """
+        # Try static evaluation first
+        try:
+            result = cast_to(BooleanLiteral, self._evaluate_expression(condition))
+        except (TypeError, ValueError, AttributeError):
```
yitchen-tim: Can we differentiate different case here? Right now the MCM case is mixed with other error cases.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -405,32 +435,34 @@ def add_gate_instruction(
...
             self._verbatim_circuit.append(CircuitInstruction(gate, target))
         else:
-            self._circuit.append(CircuitInstruction(gate, target))
+            active.append(CircuitInstruction(gate, target))
 
     def handle_parameter_value(self, value: Number | Expr) -> Number | Parameter:
         return _sympy_to_qiskit(value, self._param_map) if isinstance(value, Expr) else value
 
-    def add_measure(self, target: tuple[int], classical_targets: Iterable[int] | None = None):
+    def add_measure(self, target: tuple[int], classical_targets: Iterable[int] | None = None, **kwargs):
```
nilsquet: why was **kwargs added?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -380,13 +408,15 @@ def declare_variable(
             else:
                 size = 1
             
-            self._circuit.add_bits([Clbit() for _ in range(size)])
+            # this is used deal with Qiskit circuit storing all classical bits in a flat list
```
nilsquet:
```suggestion
            # this is used to deal with Qiskit's QuantumCircuit storing all classical bits in a flat list
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+        self._circuit_stack.pop()
+
+        # Push circuit for else-block (the interpreter always consumes this yield
+        # even for if-only blocks; the empty circuit is discarded below)
+        false_body = QuantumCircuit(main.num_qubits, main.num_clbits)
+        self._circuit_stack.append(false_body)
+        yield False
+        self._circuit_stack.pop()
+
+        # Only include false_body if it has instructions
```
nilsquet: Comment should be removed.

nilsquet (changes requested):

qiskit_braket_provider/providers/adapter.py
```diff
@@ -464,17 +496,162 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
+
+        if_else_op = IfElseOp(resolved_condition, true_body, actual_false)
+        qubits = list(range(main.num_qubits))
+        clbits = list(range(main.num_clbits))
+        main.append(if_else_op, qubits, clbits)
+
+    def evaluate_for_range(self, set_declaration, loop_var: str, loop_type):
+        """Set up each for-loop iteration, yielding once per iteration."""
+        index = self._evaluate_expression(set_declaration)
+        if isinstance(index, RangeDefinition):
```
yitchen-tim, line 599: Are we making a conscious decision that we evaluating constants on the fly? 
Should we have an unrolling pass as a separate pass? 
What if there are classical optimization that needs if-else and loop statements to be structure? What is the path it not unrolling?

qiskit_braket_provider/providers/adapter.py
aniksd-braket: This approach will not work because `_clbit_offset` tracks the starting index of each declared bit variable in the flat `clbit` list, but it doesn't distinguish between bits used as variables vs bits used for mid-circuit measurements.

aniksd-braket: But I agree with your point, let me come up with a new approach

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Good catch

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Definition has been altered now

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Changed the implementation

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Loops are out of scope for this PR. The implementation here is just to ensure non-MCM loops are handled.

nilsquet (approved):

PR #297, opened 2026-04-16
fix: enforce three execution modes in `BraketAwsBackend.run` with input validation

yitchen-tim:
**Description:**

Previously, `backend.run()` accepted arbitrary combinations of `verbatim`, `native`, `optimization_level`, and `pass_manager` with no validation. `optimization_level` defaulted to `0` and was silently ignored when `native=False`, and `pass_manager` could be used without `native=True`. This made it easy to misconfigure runs without any feedback.

This PR formalizes three execution modes and adds validation to enforce them:

1. **Default** (`verbatim=False`, `native=False`): QBP translates the circuit and passes it to the Braket service for optimization and execution.
2. **Verbatim** (`verbatim=True`): QBP translates and wraps in a verbatim box. No compilation by QBP or the Braket service.
3. **Native** (`native=True`): QBP compiles the circuit using `optimization_level` or `pass_manager`, wraps in a verbatim box, and submits. The Braket service executes without further compilation.

Validation rules:
- `verbatim` and `native` cannot both be `True`
- `native=True` requires either `optimization_level` or `pass_manager` (but not both)
- `optimization_level` and `pass_manager` require `native=True`

**Changes:**
- `optimization_level` default changed from `0` to `None` in `BraketAwsBackend.run`
- Extracted validation into `_validate_run_config` static method
- Updated docstring to document the three execution modes and parameter constraints
- Added five tests with error message matching for all validation paths
- Updated existing tests to comply with the new validation (`native=True` calls now include `optimization_level` or `pass_manager`)
- Added test verifying `optimization_level=3` triggers gate cancellation on QPU backends

**Testing:**
All 30 tests pass (4 skipped — external resource tests).


sesmart:
Actually, was wondering something similar - do you even need the guard then? Optimization_level=0 with no  coupling_map or target won't add swaps, so it really should do nothing.

nilsquet (dismissed):

aniksd-braket:
The arguments to `backend.run()` are quite confusing - if `verbatim=true` then we do not perform any transpilation, similarly if `native=false` then again we do not perform any transpilation (unless it is a simulator). So there are effectively 2 paths to the same outcome which sounds very confusing. We could eliminate one just path by having `_target_and_basis_gates` always use _target, instead of using it only for the case of a simulator. In that case, we will not need the additional `optimization_level` check that is introduced by this PR

speller26:
I'd intended to make `native=True` by default, but we decided against it in the end because compilation would happen under the hood. The transpilation not happening if the condition is false is there because there's nothing to really pass to `transpile` if there's no target or basis gates.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -389,6 +389,15 @@ def run(
...
             del options["meas_level"]
 
+        if optimization_level is not None and not native:
+            raise QiskitBraketException(
+                "optimization_level requires native=True. "
+                "Set native=True to use optimization_level."
+            )
+
+        if native and optimization_level is None:
+            raise QiskitBraketException("native=True requires optimization_level to be specified.")
```
speller26, line 399: This will break customers; the default for `native` should still be 0, as it was before.

yitchen-tim: I'd argue that `native=True` and implicitly assuming `optimization_level=0` is not a good experience. Is for very specific customers who don't want to think about qubit routing and mapping, but want to avoid gate optimization. This is the same users group for verbatim mode. And I'd say it should not be the default.

sesmart: I don't think it is so problematic, but it is a little historic and certainly could be reevaluated. `optimization_level=0` just means minimal changes are made to your circuit, and leaves the emphasis on translation rather than compilation and optimization. It makes the user specify when they want additional work done on their circuit and assumes the user wants to interact the compiler. Main inputs are indeed a verbatim-ready circuit.

Note the default for most users who don't want to touch compilation is just to leave `native=False` and let it pass through non-verbatim, which is 100% fine as well.

yitchen-tim: I updated the PR to define three clear modes to execute circuits through QBP

1. **Default** (verbatim=False, native=False) QBP does basic translation (Qiskit → Braket) and hands the circuit to the Braket service. The service handles optimization, routing, and mapping before execution.
2. **Verbatim** (verbatim=True) QBP just translates it to Braket format, wraps it in a verbatim box, and submits it. No optimization, routing, or mapping happens — not by QBP, not by the Braket service.
3. **Native** (native=True) QBP performs compilation. It optimizes, routes, and maps the circuit using your optimization_level or pass_manager. Then it wraps the result in a verbatim box and submits it. The Braket service runs it as-is, no further compilation.

Validation is to ensure users does not accidentally fall out of these modes, and prevent specifying conflicting config parameters that would lead to unexpected behavior.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -352,30 +352,46 @@ def run(
...
-        optimization_level: int = 0,
+        optimization_level: int | None = None,
         callback: Callable | None = None,
         num_processes: int | None = None,
         pass_manager: PassManager | None = None,
         **options,
     ):
         """Execute ``QuantumCircuit``s on a ``BraketAwsBackend``
 
+        Three modes of execution are supported by Qiskit-Braket-Provider (QBP):
```
sesmart, line 360: I really like this by the way - it's just helpful as a quick reminder 😁

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -424,6 +442,56 @@ def _target_and_basis_gates(
...
+        if native and optimization_level is not None and pass_manager is not None:
+            raise QiskitBraketException(
+                "optimization_level and pass_manager cannot both be specified. "
+                "Provide one or the other when using native=True."
+            )
+
+        if native and optimization_level is None and pass_manager is None:
+            raise QiskitBraketException(
+                "native=True requires either optimization_level or pass_manager to be specified."
+            )
```
speller26, line 484: I'm still not sold on requiring either `optimization_level` or `pass_manager` for native execution; I would prefer the optimization level to default to Qiskit's instead.

As well, this breaks almost all of the [idiomatic examples](https://github.com/amazon-braket/qiskit-braket-provider/blob/main/docs/tutorials/5_tutorial_transpilation.ipynb).

aniksd-braket: I agree with @speller26. If `native=True` I think we should fall back to Qiskit's default optimization level which I think is 2.

PR #298, opened 2026-04-16
Add minimum dependency version CI

nilsquet:
## Summary

Adds a min-versions tox environment that installs all direct dependencies at their lowest declared versions and runs the full test suite, ensuring the package actually works at its stated compatibility floor.

## Changes

requirements.txt
- qiskit>=0.34.2 → >=2.0.0 (required for BoxOp, BackendEstimatorV2, BackendSamplerV2)
- qiskit-ionq>=0.5.2 → >=1.0.0 (first release compatible with qiskit 2.0)
- amazon-braket-sdk>=1.99.0 → >=1.111.0 (required for Circuit.barrier() and OpenQASM interpreter fixes)

requirements-dev.txt
- qiskit-algorithms>=0.2.2 → >=0.4.0 (compatible with qiskit 2.0)
- coverage>=5.5 → >=7.6.1 (required for coverage.types.Tracer used by numba)
- pytest>=6.2.5 → >=7.0.0 (required for _pytest.scope used by anyio)
- ruff → >=0.4.0 (pin a minimum to avoid unpinned lowest-direct resolution picking up ancient versions)
- Removed unused mypy and mypy-extensions


requirements.txt
```diff
@@ -1,7 +1,7 @@
 certifi>=2021.5.30
-qiskit>=0.34.2
-qiskit-ionq>=0.5.2
-amazon-braket-sdk>=1.99.0
+qiskit>=2.0.0
```
rmshaffer, line 2: note there's also a Qiskit badge in README.md which should be updated accordingly to match this `>=2.0.0` constraint.

https://github.com/qiskit-community/qiskit-braket-provider/blob/4c888a61b642b75badbc65f77c48a31a8541bc0f/README.md?plain=1#L5

aniksd-braket:
Do we still need the minimum version tests if we have decided to bump up the minimum version to the one that would work across all tests?

nilsquet:
> Do we still need the minimum version tests if we have decided to bump up the minimum version to the one that would work across all tests?

Yes, because it allows us to check against the lowest of all our dependencies that are supposed to work. This would would have also detected the originally missing lower bound in this [PR](https://github.com/qiskit-community/qiskit-braket-provider/pull/295).

requirements.txt
nilsquet: Good catch, I updated it.

aniksd-braket (dismissed):

requirements-dev.txt
```diff
@@ -1,14 +1,12 @@
...
-mypy-extensions>=0.4.3
+pytest>=7.0.0
 tox==3.24.5
 jupyter-sphinx>=0.3.2
 nbsphinx>=0.8.8
 sphinx-autodoc-typehints>=1.17.0
 qiskit_sphinx_theme~=1.16.0
 reno>=3.5.0
-ruff
+ruff>=0.4.0
```
speller26, line 12: nit: don't think this is necessary

speller26 (dismissed):

speller26 (approved):

PR #307, opened 2026-04-21
feature: adding loop support for OpenQASM

aniksd-braket:
### Description

Adds support for translating OpenQASM for and while loops into Qiskit's native `ForLoopOp` and `WhileLoopOp` control flow operations, instead of  statically unrolling them.

For loops are captured into a `ForLoopOp` by pushing a body circuit onto the stack, yielding once for the interpreter to populate it, then wrapping it with the loop's index set.

While loops with mid-circuit measurement (MCM) conditions are captured into a `WhileLoopOp` using the same pattern. Static while loops (no MCM dependency) continue to evaluate the condition directly.

Break/continue statements in loops raise `NotImplementedError`, since Qiskit's `ForLoopOp/WhileLoopOp` do not support these constructs. Detection relies on `handle_loop_break()` and `handle_loop_continue()` hooks added in [amazon-braket-default-simulator-python#362]

### Changes
- **adapter.py**: Added `ForLoopOp`, `WhileLoopOp` imports; rewrote `evaluate_for_range` to produce `ForLoopOp`; added MCM path to `evaluate_while_condition` to produce WhileLoopOp; added `handle_loop_break` and `handle_loop_continue` overrides
- **test_adapter_branching.py**: Added/updated tests for for-loop and while-loop translation, break/continue rejection, circuit dimension expansion, and unsupported condition types
- **Empty body validation** — MCM-conditioned loops and branches with no quantum operations in the body raise `ValueError`.

### Dependencies
- Requires [amazon-braket-default-simulator-python feature v1.38.1] for the `handle_loop_break/handle_loop_continue` hooks


qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
         main.append(if_else_op, qubits, clbits)
 
     def evaluate_for_range(self, set_declaration, loop_var: str, loop_type):
-        """Set up each for-loop iteration, yielding once per iteration."""
+        """Capture the for-loop body into a ForLoopOp.
+
+        Yields once to capture the body. If the body contains quantum operations,
+        wraps it in a ForLoopOp. If purely classical (empty body circuit),
```
yitchen-tim, line 614: Just a question and not requesting a change: Is there a gap in Qiskit that we cannot construct classical expression? Ideally, we should move to a direction that we don't resolve value on the fly, but instead, have a pass for that. Is the on-the-fly resolution for incremental build or for Qiskit feature gap?

yitchen-tim (dismissed):

qiskit_braket_provider/providers/adapter.py
speller26: I think there's some rudimentary support for classical variables in Qiskit, but not much. However, for this implementation, I'm in favor of keeping the current behavior because
1. Devices don't currently have a way to signal that they support MCM, and translating purely classical control flow to Qiskit control flow will break this functionality for current devices
2. It doesn't require any extra implementation; purely classical control flow has been supported out of the box since we introduced the Qiskit context.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
-                self.declare_variable(loop_var, loop_type, i)
-                yield
+
+        main = self._active_circuit
+        body = QuantumCircuit(main.num_qubits, main.num_clbits)
+        self._circuit_stack.append(body)
+        with self.enter_scope():
+            self.declare_variable(loop_var, loop_type, index_values[0])
+            yield
+        self._circuit_stack.pop()
```
speller26: Love to see it in action

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
         if isinstance(index, RangeDefinition):
             index_values = [IntegerLiteral(x) for x in convert_range_def_to_range(index)]
         else:
             index_values = index.values
-        for i in index_values:
-            with self.enter_scope():
-                self.declare_variable(loop_var, loop_type, i)
-                yield
+
+        main = self._active_circuit
```
speller26, line 626: nit: is `_active_circuit` really necessary? Why not just `self._circuit_stack[-1]`?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
+            yield
+        self._circuit_stack.pop()
+
+        if not body.data:
+            # Purely classical loop body — statically unroll remaining iterations
+            for i in index_values[1:]:
+                with self.enter_scope():
+                    self.declare_variable(loop_var, loop_type, i)
+                    yield
+            return
```
speller26: Ah, so there is some extra implementation; I don't want this to be necessary for purely classical control flow, and I'll fix up the default simulator accordingly.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
-        for i in index_values:
-            with self.enter_scope():
-                self.declare_variable(loop_var, loop_type, i)
-                yield
+
+        main = self._active_circuit
+        body = QuantumCircuit(main.num_qubits, main.num_clbits)
+        self._circuit_stack.append(body)
+        with self.enter_scope():
+            self.declare_variable(loop_var, loop_type, index_values[0])
```
nilsquet, line 633: When capturing the body, `i` is declared as the first iteration value (`index_values[0]`), so expressions using `i` are resolved to that literal at capture time. The `Parameter("i")` passed to `ForLoopOp` is a separate symbol that never appears in `body.parameters`, so Qiskit's per-iteration rebinding is a no-op.

Result: `ForLoopOp` stores the `i == 0` specialization, and Qiskit re-runs that same body for every iteration.

**Example**

```qasm
for int[8] i in [0:2] {
    rx(i * 0.5) q[0];
}
```

- Expected: `rx(0.0); rx(0.5); rx(1.0)`
- Actual: `body = [rx(0.0) q[0]]`, `body.parameters == []`, so `rx(0.0)` runs three times.

**Why tests don't catch it**
None of the new for-loop tests reference the loop variable in the body, so the bug is invisible to the test suite. A test like the following would catch it:

```python
def test_for_loop_body_references_loop_var():
    qasm = """
    OPENQASM 3.0;
    qubit[1] q;
    for int[8] i in [0:2] { rx(i * 0.5) q[0]; }
    """
    qc = to_qiskit(qasm)
    for_op = next(d.operation for d in qc.data if isinstance(d.operation, ForLoopOp))
    _, loop_param, body = for_op.params
    # The body must reference loop_param symbolically, not the literal 0.
    assert loop_param in body.parameters
```

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
+        max_qubits = max(body.num_qubits, main.num_qubits)
+        max_clbits = max(body.num_clbits, main.num_clbits)
+        if max_qubits > main.num_qubits:
+            main.add_bits([Qubit() for _ in range(max_qubits - main.num_qubits)])
+        if max_clbits > main.num_clbits:
+            main.add_bits([Clbit() for _ in range(max_clbits - main.num_clbits)])
+        main.append(for_op, list(range(max_qubits)), list(range(max_clbits)))
+
+    def handle_loop_break(self):
+        raise NotImplementedError("break statements are not supported in loops.")
```
nilsquet, line 680: Nit: docstring would be great

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
+
+    def handle_loop_continue(self):
+        raise NotImplementedError("continue statements are not supported in loops.")
```
nilsquet, line 684: Nit: docstring would be great

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
         main.append(if_else_op, qubits, clbits)
 
     def evaluate_for_range(self, set_declaration, loop_var: str, loop_type):
```
nilsquet: Nit: would you mind adding types for the parameters? I know that you didn't change this line, it would just be a another small step to increase our code quality.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -597,21 +605,98 @@ def evaluate_condition(self, condition):
...
+            main.add_bits([Clbit() for _ in range(max_clbits - main.num_clbits)])
+        main.append(for_op, list(range(max_qubits)), list(range(max_clbits)))
+
+    def handle_loop_break(self):
+        raise NotImplementedError("break statements are not supported in loops.")
+
+    def handle_loop_continue(self):
+        raise NotImplementedError("continue statements are not supported in loops.")
 
     def evaluate_while_condition(self, condition):
```
nilsquet: Nit: could you add a type for `condition`? I know that you didn't change this line, it would just be a another small step to increase our code quality.

tests/providers/test_adapter_dynamic_circuits.py
```diff
@@ -1,7 +1,7 @@
 """Tests for branching statement (if/else) support in the Qiskit adapter."""
```
nilsquet: The docstring should be expanded to cover loops as well.

speller26:
Also, can you bump the version to 0.14.0?

nilsquet (changes requested):

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Qiskit does not have expansive support for classical expressions. There is [this](https://quantum.cloud.ibm.com/docs/en/api/qiskit/circuit_classical), but it is quite minimal and exploratory. 

> Is the on-the-fly resolution for incremental build or for Qiskit feature gap?

For the above, I would say that the answer is both.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: referencing `self_circuit_stack[-1]` just felt a little fragile across the class. Using a private property seemed like a cleaner approach

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Yup, because a purely classical for loop cannot be passed to Qiskit's `ForLoopsOp`, as the body does not contain a quantum circuit

qiskit_braket_provider/providers/adapter.py
speller26: No strong opinion here; my thought was `self_circuit_stack[-1]` clarifies the purpose of the stack.

qiskit_braket_provider/providers/adapter.py
aniksd-braket: Good catch, will handle this case

nilsquet (dismissed):
The symbolic binding fix looks correct — nice work! Three nits:

**Nit: `q[i]` gives an unhelpful error**

When the loop variable is used as a qubit index, the symbolic recapture crashes deep in the simulator with `AttributeError: 'SymbolLiteral' object has no attribute 'values'`. Could we catch this early and raise something like `"Loop variable as qubit index is not supported — qubit indices must be statically known"`?

```python
to_qiskit("""
OPENQASM 3.0;
qubit[3] q;
for int[8] i in [0:3] {
    h q[i];
}
""")
```

**Nit: mixed classical + quantum loop body should error**

If the body has both quantum ops and classical side effects (e.g. variable assignment), the classical mutations won't be captured by `ForLoopOp`. Currently this silently gives wrong results. Could we detect this and raise an error?

```python
to_qiskit("""
OPENQASM 3.0;
qubit[1] q;
int[8] counter = 0;
for int[8] i in [0:3] {
    counter = counter + 1;
    rx(i * 0.5) q[0];
}
rx(counter * 0.1) q[0];
""")
# Silently produces rx(0.2) instead of rx(0.4)
```

**Nit: empty range crashes**

`index_values[0]` blows up with `IndexError` when the range is empty. Could use a guard to return early.

```python
to_qiskit("""
OPENQASM 3.0;
qubit[1] q;
for int[8] i in [0:-1] {
    h q[0];
}
""")
```

rmshaffer:
> Also, can you bump the version to 0.14.0?

Just a side note: the new release pipeline (which didn't yet exist when @speller26 made this request originally 😄) will automatically bump the minor version if there is a commit with the prefix `feature:`. So this should happen automatically, no need to manually bump versions anymore.

nilsquet (approved):

PR #310, opened 2026-04-22
fix: Ignore barriers if unsupported

### LINKED ISSUE #300: `measure_all` causes ValidationException on Braket devices that don't support barrier
## Summary

PR #275 (shipped in v0.11.1) added barrier translation in `_translate_to_braket`, converting Qiskit `Barrier` instructions to `braket_circuit.barrier(...)`. This causes a `ValidationException` when submitting to any Braket QPU or simulator because gate-model `barrier` is not supported in OpenQASM 3.0 on any Braket device.

## Impact

`QuantumCircuit.measure_all()` in Qiskit implicitly inserts a barrier across all qubits before the measurements. Any circuit using `measure_all()` — which is extremely common — will **silently break** after upgrading from v0.10.0 to v0.11.x.

## Steps to Reproduce

```python
from qiskit import QuantumCircuit
from qiskit_braket_provider import AWSBraketProvider

circuit = QuantumCircuit(3, 3)
circuit.name = "3-qubit GHZ"
circuit.h(0)
circuit.cx(0, 1)
circuit.cx(1, 2)
circuit.measure_all()  # implicitly adds: barrier q[0], q[1], q[2];

# circuit now contains:
# OPENQASM 3.0;
# qubit[3] q;
# h q[0];
# cnot q[0], q[1];
# cnot q[1], q[2];
# barrier q[0], q[1], q[2];   <-- inserted by measure_all()
# b[0] = measure q[0];
# b[1] = measure q[1];
# b[2] = measure q[2];

provider = AWSBraketProvider()
backend = provider.get_backend("sv1")
job = backend.run(circuit, shots=10)
```

**Error:**
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the
CreateQuantumTask operation: [line 7] Uses 'barrier' which is not supported on device
```

## Root Cause

In v0.10.0, `_translate_to_braket` silently dropped barrier instructions (no `case "barrier"` handler). PR #275 added:

```python
case "barrier":
    qubit_indices = [qubit_labels[circuit.find_bit(qubit).index] for qubit in qubits]
    braket_circuit.barrier(target=qubit_indices if qubit_indices else None)
```

This correctly translates Qiskit barriers to Braket barriers, which are then serialized to `barrier q[0], q[1], q[2];` in the OpenQASM 3.0 action submitted to AWS. However, AWS Braket rejects this instruction on all QPU and simulator devices at the API level.

Note: Braket does support `barrier` in **pulse-level** (`defcal`) programs for Rigetti devices — which is why `barrier` appears in Rigetti's `nativeGateSet` — but this is unrelated to gate-model circuit barriers in OpenQASM 3.0.

sesmart:
### Summary

Barrier introduced a breaking change for non-barrier supporting devices, documented in #300. Whereas barrier was previously globally ignored, now it is allowed for submission to Braket, and fails when not supported. 

Here we revert that change for non-supporting Barriers, removing all barriers for non-supporting devices and raising a warning. Closes #300. 

### Details and comments

- Requires Barrier be added to Target for simulators, similar to QPUs.
- Raises a warning if barrier instructions are present but being ignored.


qiskit_braket_provider/providers/adapter.py
```diff
@@ -1203,6 +1204,27 @@ def _restore_verbatim_boxes(
...
+    circ: QuantumCircuit, verbatim_box_name: str = _BRAKET_VERBATIM_BOX_NAME
+):
+    """If the instruction before a final measurements is a non-verbatim barrier, remove it.
+
+    Main application is to help remove the terminal barrier added by (1) measure_all, or
+    (2) measure_active in cases where the device does not support barrier
+    [TODO: refactor as a TransformationPass]
+    """
+    n_i = len(circ.data)
+    for n, instr in enumerate(circ.data[::-1]):
```
yitchen-tim: You should also break if it see an instruction other than measure or barrier, other it removes all barrier

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1681,6 +1706,7 @@ def _default_target(circuits: Iterable[QuantumCircuit]) -> Target:
         if name := _BRAKET_TO_QISKIT_NAMES.get(braket_name.lower()):
             target.add_instruction(instruction, name=name)
     target.add_instruction(Measure())
+    target.add_instruction(Barrier(1))
```
yitchen-tim, line 1811: For uneducated me: (1) why is it needed? (2) Is single qubit barrier enough? I think measure_all adds multiple-qubit barrier.

test/unit_tests/test_adapter.py
```diff
@@ -700,6 +700,27 @@ def test_barrier_single_qubit(self):
...
+
+        expected = Circuit().h(0).cnot(0, 1).measure([0, 1, 2])
+        self.assertEqual(
+            to_braket(
+                qc,
+                target=target,
+            ),
+            expected,
+        )
+
```
yitchen-tim, line 735: Also add tests to make sure non-terminal barrier is not removed

qiskit_braket_provider/providers/adapter.py
sesmart: This follows the previous PR - it doesn't affect multi-qubit Barriers at all, as the target in the transpiler treats barriers as compilation instructions, and we have a test for it to make sure multi-qubit barriers are ok with 1Q targets. You also can't provide a variable qubit Barrier instruction. 

Specifically, we need this because (1) we support barrier instructions in our simulators, but (2) they, unlike QHPs, do not list them as supported gates in the action or paradigm.

qiskit_braket_provider/providers/adapter.py
sesmart: Ah, right. I had removed that but yes, there should be an additional break once another instruction is encountered.

test/unit_tests/test_adapter.py
sesmart: Done.

qiskit_braket_provider/providers/adapter.py
sesmart: It is added now, feel free to resolve if acceptable.

yitchen-tim: Is this correct? Won't the last `break` just breaks the loop after 1 iteration? Shouldn't it be 
```
if Measure:
    ....
elif Barrier:
    ....
else:
    break
```

sesmart: It looks weird, but this is trying to be as selective as possible - i.e. there is only 1 scenario where Barrier is removed. 

1) measure must be the first set of instructions; after each measure we `continue` to next instruction
2) the next non-measure instruction must be a barrier -> if so, remove it and `break`
3) if any other gate is encountered, simply `break`

This tries to capture the .measure_all() or .measure_active() logic as closely as possible, where we deterministically add a barrier + measure set of instructions.

yitchen-tim: I see, is this because `measure_all` insert `measure([1,2,3])` instead of three measurements, so there is only one measurement?

yitchen-tim (dismissed):

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1321,6 +1322,28 @@ def _restore_verbatim_boxes(
     return reconstructed_circuit
 
 
+def _remove_terminal_barrier(
```
speller26: I generally don't like performing surgery on a created object (in this case a `QuantumCircuit`), but I'm happy to merge this if others are fine with it. My preference would be to, if `barrier` is not supported by the device (and in turn target/basis_gates), track barriers as we're translating the circuit, skip adding the barrier until we know it's not terminal.

In fact, can we lean on the device properties for barrier support as a whole?

sesmart: Okay, changed the logic here - now the check is simply in the translation, and occurs based on the Target. All non supported barriers are removed and a warning is raised.

speller26 (approved):

PR #321, opened 2026-04-28
infra: add type hint and check via mypy to CI

nilsquet:
*Issue #, if available:*

*Description of changes:*
- Add type hints to QBP
- Add mypy CI run
- Make sure mypy runs succesfully

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/docs/tutorials/2_tutorial_hybrid_jobs.ipynb
+++ b/docs/tutorials/2_tutorial_hybrid_jobs.ipynb
@@ -66,7 +66,7 @@
    "outputs": [],
    "source": [
     "@hybrid_job(device=sv1.arn, include_modules=\"qiskit_algorithms\")\n",
-    "def main() -> None:\n",
+    "def main():\n",
     "    \"\"\"Decorated function that will be run in the docker container.\"\"\"\n",
     "    backend = BraketProvider().get_backend(\"SV1\")\n",
     "\n",
--- a/docs/tutorials/2_tutorial_hybrid_jobs.ipynb
+++ b/docs/tutorials/2_tutorial_hybrid_jobs.ipynb
@@ -66,7 +66,7 @@
    "outputs": [],
    "source": [
     "@hybrid_job(device=sv1.arn, include_modules=\"qiskit_algorithms\")\n",
-    "def main():\n",
+    "def main() -> None:\n",
     "    \"\"\"Decorated function that will be run in the docker container.\"\"\"\n",
     "    backend = BraketProvider().get_backend(\"SV1\")\n",
     "\n",
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -7,7 +7,7 @@ import enum
 import logging
 import warnings
 from abc import ABC
-from typing import TYPE_CHECKING, Any, Generic, NoReturn, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
 from qiskit import QuantumCircuit
 from qiskit.providers import BackendV2, Options, QubitProperties
@@ -152,16 +152,16 @@ class BraketLocalBackend(BraketBackend[LocalSimulator]):
     def qubit_properties(self, qubit: int | list[int]) -> QubitProperties | list[QubitProperties]:
         raise NotImplementedError
 
-    def drive_channel(self, qubit: int) -> NoReturn:
+    def drive_channel(self, qubit: int):
         raise NotImplementedError(f"Drive channel is not supported by {self.name}.")
 
-    def measure_channel(self, qubit: int) -> NoReturn:
+    def measure_channel(self, qubit: int):
         raise NotImplementedError(f"Measure channel is not supported by {self.name}.")
 
-    def acquire_channel(self, qubit: int) -> NoReturn:
+    def acquire_channel(self, qubit: int):
         raise NotImplementedError(f"Acquire channel is not supported by {self.name}.")
 
-    def control_channel(self, qubits: Iterable[int]) -> NoReturn:
+    def control_channel(self, qubits: Iterable[int]):
         raise NotImplementedError(f"Control channel is not supported by {self.name}.")
 
     def run(
@@ -345,16 +345,16 @@ class BraketAwsBackend(BraketBackend[AwsDevice]):
     def meas_map(self) -> list[list[int]]:
         raise NotImplementedError(f"Measurement map is not supported by {self.name}.")
 
-    def drive_channel(self, qubit: int) -> NoReturn:
+    def drive_channel(self, qubit: int):
         raise NotImplementedError(f"Drive channel is not supported by {self.name}.")
 
-    def measure_channel(self, qubit: int) -> NoReturn:
+    def measure_channel(self, qubit: int):
         raise NotImplementedError(f"Measure channel is not supported by {self.name}.")
 
-    def acquire_channel(self, qubit: int) -> NoReturn:
+    def acquire_channel(self, qubit: int):
         raise NotImplementedError(f"Acquire channel is not supported by {self.name}.")
 
-    def control_channel(self, qubits: Iterable[int]) -> NoReturn:
+    def control_channel(self, qubits: Iterable[int]):
         raise NotImplementedError(f"Control channel is not supported by {self.name}.")
 
     def run(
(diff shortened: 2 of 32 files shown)


docs/tutorials/2_tutorial_hybrid_jobs.ipynb
```diff
@@ -66,7 +66,7 @@
    "outputs": [],
    "source": [
     "@hybrid_job(device=sv1.arn, include_modules=\"qiskit_algorithms\")\n",
-    "def main():\n",
+    "def main() -> None:\n",
```
laurencap: Nit: I think it's probably better not to include this in a tutorial, feels distracting

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -33,6 +33,12 @@
 )
 from .braket_quantum_task import BraketQuantumTask
 
+if TYPE_CHECKING:
```
laurencap, line 36: Why?

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -43,14 +49,14 @@
 class BraketBackend(BackendV2, ABC, Generic[T]):
     """Base Qiskit backend for Amazon Braket devices."""
 
-    def __init__(self, device: T, name: str, **fields):
+    def __init__(self, device: T, name: str, **fields: Any) -> None:  # noqa: ANN401
```
laurencap: Adding this caused a new style failure?

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -85,7 +91,12 @@ def get_gateset(self, native: bool = False) -> set[str] | None:
         action = self._device.properties.action.get(DeviceActionType.OPENQASM)
         return gateset_from_properties(action) if action else None
 
-    def _run_program_set(self, braket_circuits: list[Circuit], shots: int | None, **options):
+    def _run_program_set(
+        self,
+        braket_circuits: list[Circuit],
+        shots: int | None,
+        **options: Any,  # noqa: ANN401
```
laurencap: Same question here

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -141,20 +152,24 @@ def meas_map(self) -> list[list[int]]:
     def qubit_properties(self, qubit: int | list[int]) -> QubitProperties | list[QubitProperties]:
         raise NotImplementedError
 
-    def drive_channel(self, qubit: int):
+    def drive_channel(self, qubit: int) -> NoReturn:
```
laurencap: I feel like `NoReturn` is actually unhelpful here. It's not implemented, but for an interface / ABC, we want to know what the return type should be when implemented. Better to leave blank than include this imo

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -49,8 +54,8 @@ def __init__(
         *,
         verbatim: bool = False,
         optimization_level: int = 0,
-        **options,
-    ):
+        **options: Any,  # noqa: ANN401
```
laurencap: Having these everywhere `Any,  # noqa: ANN401` seems counterproductive when **name at the end of a func signature always means kwargs. Also, shouldn't this be `dict[str, Any]`?

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -301,11 +314,13 @@ def test_to_braket_backward_compatibility():
...
 )
-def test_to_braket_verbatim_and_layout_options(verbatim, layout_method, h_cx_circuit):
+def test_to_braket_verbatim_and_layout_options(
+    verbatim: bool, layout_method: str | None, h_cx_circuit: QuantumCircuit
+) -> None:
     qc = QuantumCircuit(NUM_QUBITS)
     qc.append(BoxOp(h_cx_circuit, label=VERBATIM_LABEL), QUBIT_PAIR)
 
-    kwargs = {"verbatim": verbatim}
+    kwargs: dict[str, Any] = {"verbatim": verbatim}
```
laurencap: Not necessary, remove this

test/unit_tests/test_adapter_to_qiskit_verbatim.py
```diff
@@ -158,7 +163,7 @@ def test_verbatim_box_with_measurements():
     assert len(measurements) == 2
 
 
-def test_verbatim_box_qubit_mapping():
+def test_verbatim_box_qubit_mapping() -> None:
```
laurencap: Remove all the `-> None` from the test files. Not helpful, not idiomatic

laurencap (changes requested):
This is awesome. I requested some changes to keep things idiomatic and make sure that the type hints that we include are providing useful information and aren't noisy

docs/tutorials/2_tutorial_hybrid_jobs.ipynb
nilsquet: Enabled ruff's `suppress-none-returning` in `pyproject.toml` so ANN201 no longer demands return annotations on functions that don't return anything. Reverted `def main() -> None:` here and `-> None` on the plotting helper in tutorial 4. Kept `-> list[SolutionSample]` in tutorial 3 since it documents the return type for a reader.

qiskit_braket_provider/providers/braket_backend.py
nilsquet: Quick background: `TYPE_CHECKING` is `False` at runtime but `True` for type checkers, so imports inside only load during static analysis. Combined with `from __future__ import annotations`, this lets us reference types in annotations without the runtime cost or circular-import risk.

Two reasons this block exists:
1. `datetime`, `Callable`, `Iterable` — only used in annotations. Ruff's `TC003` rule requires annotation-only stdlib imports under `TYPE_CHECKING` when `from __future__ import annotations` is present.
2. `BraketProvider` — genuine circular: `braket_provider.py` imports `BraketAwsBackend`/`BraketLocalBackend` from this module. Runtime-importing `BraketProvider` here would loop.

qiskit_braket_provider/providers/braket_backend.py
nilsquet: The noqa was treating a symptom: `ANN003` demanded an annotation on `**kwargs`, the only honest annotation was `Any`, which tripped `ANN401`, hence the noqa. Disabled `ANN003` (and `ANN002` for `*args`) in `pyproject.toml`. 27 `# noqa: ANN401` comments that existed solely to suppress the `Any` fallout are gone.

Two legitimate `Any` usages still carry inline `# noqa: ANN401`:
- `value: Any = None` in `declare_variable` — matches the base class; OpenQASM literals are genuinely dynamic.
- `_evaluate_expression(...) -> Any` — wide union return; could be tightened in a follow-up.

qiskit_braket_provider/providers/braket_backend.py
nilsquet: Enabled `suppress-none-returning = true` under `[tool.ruff.lint.flake8-annotations]`. ANN201 now skips functions with no `return` statement.

Same setting resolves your test-file comment below (removed `-> None` from all 288 test functions) without a per-file ignore.

qiskit_braket_provider/providers/braket_estimator.py
nilsquet: Resolved by the comment 3 fix: `**options` has no annotation or noqa now.

On `dict[str, Any]`: per PEP 484 ("Arbitrary argument lists"), `**kwargs` is annotated with the *value* type, and inside the function `kwargs` is deduced as `dict[str, <value-type>]`. So `**kwargs: Any` already means `kwargs: dict[str, Any]` at runtime.

test/unit_tests/test_adapter_to_braket_verbatim.py
nilsquet: Done. Also inlined the call:`layout_method` already accepts `None`, so the conditional kwargs dict collapses to `bc = to_braket(qc, verbatim=verbatim, layout_method=layout_method)`.

test/unit_tests/test_adapter_to_qiskit_verbatim.py
nilsquet: Done.

rmshaffer (approved):

aniksd-braket (approved):

laurencap (approved):

PR #322, opened 2026-04-28
fix: missing layout bug

aniksd-braket:
Fix a bug where the circuit layout was being lost during `_restore_verbatim_boxes()`

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -1275,6 +1275,7 @@ def _restore_verbatim_boxes(
         transpiled_circuit.num_qubits, transpiled_circuit.num_clbits
     )
     reconstructed_circuit.global_phase = transpiled_circuit.global_phase
+    reconstructed_circuit._layout = transpiled_circuit._layout
 
     verbatim_box_iter = iter(verbatim_boxes)
     barrier_count = 0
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/adapter.py
```diff
@@ -1275,6 +1275,7 @@ def _restore_verbatim_boxes(
         transpiled_circuit.num_qubits, transpiled_circuit.num_clbits
     )
     reconstructed_circuit.global_phase = transpiled_circuit.global_phase
+    reconstructed_circuit._layout = transpiled_circuit._layout
```
speller26: Is it possible to set this without just directly changing the private property?

speller26 (dismissed):

test/unit_tests/test_adapter.py
```diff
@@ -1761,3 +1761,25 @@ def test_empty_circuit_with_barrier(self):
         with self.assertRaises(ValueError) as context:
             to_braket(circuit)
         self.assertIn("Cannot add global barrier to empty circuit", str(context.exception))
+
+    def test_compile_preserves_layout_with_verbatim_boxes(self):
+        """Layout from transpilation should be preserved after restoring verbatim boxes."""
+        from qiskit.circuit.library import CXGate, HGate
```
sesmart: (you also can just use qiskit_gates here without importing again)

test/unit_tests/test_adapter.py
```diff
@@ -1761,3 +1761,25 @@ def test_empty_circuit_with_barrier(self):
...
+        from qiskit.transpiler import InstructionProperties, Target
+
+        from qiskit_braket_provider.providers.adapter import _compile
```
nilsquet: Imports should go to the top.

speller26 (approved):

Issue #325, opened 2026-05-11
Allow for device emulators 


sesmart:
This issue is part of **unitaryHACK26**. You have to be [registered](https://unitaryhack.dev/register/) to complete this issue.
Learn more about the PR submission process [here](https://unitaryhack.dev/hacker-guide/) and about unitaryHACK rules [here](https://unitary.design/rules/)!

**A note about AI Slop:** while we are open to collaboration with LLMs for unitaryHACK, fully AI-generated PRs are not acceptable. It is at the maintainers' discretion whether or not LLM-generated PRs are the right fit for the issues, and those that appear fully AI-generated may be immediately rejected. Read unitaryHACK's full AI Policy [here](https://unitaryhack.dev/ai-guide/) 

---

### What is the expected behavior?

To enable support for Amazon Braket SDK Emulators, which are accessible via `device.emulator()`. Emulators do not have device properties, and so cannot be cast directly into targets. Thus, we will create them directly in the Braket Backends, using an additional property.  

### What does the solution look like? 

1. Implement a `device.emulator()` method in `BraketAwsDevice`, akin to the Braket devices. 
2. Allow for an `emulator = True` option in `provider.get_backend()`
3. Type `emulator` should be a `BraketLocalDevice`
4. Should have testing, and works with Braket primitives seamlessly. 
5. Optionally, returns some metadata that the device is an emulator.

barryPdunham:
I’m working on this issue and have a local prototype that appears to address the main emulator-routing path.

Current approach:

* add an `emulator=True` option through `BraketProvider.backends()` / `get_backend()`
* preserve the original `AwsDevice` for metadata, target construction, properties, and topology
* add a separate execution device so `device.emulator()` is used for task submission when requested
* route backend execution plus Braket Sampler/Estimator primitive execution through that execution device
* keep normal AWS-device behavior unchanged

Local validation before rebuilding my Codespace/fork workflow:

* full unit suite: `321 passed, 4 skipped`
* `ruff check`: passed
* `tox -e mypy`: passed

I’m rebuilding the branch cleanly from my fork before opening a PR, but wanted to note that this is in progress so I don’t duplicate anyone else’s work.

barryPdunham:
Opened PR #338 for this: https://github.com/amazon-braket/qiskit-braket-provider/pull/338

sesmart:
@zkasuran - can you comment on this issue?

zkasuran:
[...]

PR #333, opened 2026-06-04
feat: add device emulator support

zkasuran:
*Issue #, if available:* #325

*Description of changes:*

Adds support for Amazon Braket device emulators to the Qiskit provider. Emulators run circuits locally while mimicking a real device's gate set, connectivity and noise, so a user can check that a program will run on a target device without submitting tasks to the Braket service.

`BraketEmulatorBackend` derives from `BraketLocalBackend`: an emulator is a local execution device, so it shares the local run path. The differences are in construction. Emulators expose no device properties, so the `Target` is built from the source `AwsDevice` while circuits run on `device.emulator()`. And `shots=0` raises with a pointer to `BraketLocalBackend`, because the emulator reproduces QPU behavior and cannot return a state vector.

- `BraketAwsBackend.emulator` property returns a `BraketEmulatorBackend` for the device (issue point 1)
- `BraketProvider.get_backend(name, emulator=True)` and `backends(..., emulator=True)` (issue point 2); managed simulators are excluded since only QPUs have emulators
- `BraketEmulatorBackend` is a `BraketLocalBackend` (issue point 3)
- works with Qiskit primitives, covered by a `BackendSamplerV2` test (issue point 4)
- the `emulator` property returns `True` as metadata that the backend is an emulator (issue point 5)

AI disclosure: this change was implemented with substantial help from Claude (Anthropic), which drafted the implementation and tests. I reviewed the design and verified it locally with tox before submitting; results below.

*Testing done:*

The emulator tests build a real `LocalEmulator` from mock device capabilities and execute circuits on it, with no AWS calls or account config. They cover the `.emulator` property, `get_backend(emulator=True)` with the QPU-only filter, an emulated run, the `shots=0` error and `BackendSamplerV2` integration.

- `tox`: 325 passed, 4 skipped; coverage 100%; ruff format and check clean
- `tox -e mypy`: clean

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -18,6 +18,7 @@ from braket.aws.queue_information import QueueDepthInfo
 from braket.circuits import Circuit
 from braket.device_schema import DeviceActionType
 from braket.devices import Device, LocalSimulator
+from braket.emulation.emulator import Emulator
 from braket.program_sets import ProgramSet
 from braket.tasks.local_quantum_task import LocalQuantumTask
 
@@ -208,6 +209,168 @@ class BraketLocalBackend(BraketBackend[LocalSimulator]):
         )
 
 
+class BraketEmulatorBackend(BraketBackend[Emulator]):
+    """Runs quantum circuits on a local emulator of an Amazon Braket device.
+
+    A Braket :class:`~braket.emulation.emulator.Emulator` mimics the gate set,
+    connectivity and noise of a real ``AwsDevice`` while executing locally, without
+    submitting tasks to the Amazon Braket service. Emulators do not expose device
+    properties and therefore cannot be turned into a Qiskit ``Target`` directly, so
+    the ``Target`` is built from the source ``AwsDevice`` while circuits are executed
+    on the wrapped emulator.
+    """
+
+    def __init__(
+        self,
+        device: AwsDevice,
+        provider: BraketProvider | None = None,
+        name: str | None = None,
+        description: str | None = None,
+        online_date: datetime.datetime | None = None,
+        backend_version: str | None = None,
+        **fields,
+    ) -> None:
+        """Initialize the backend.
+
+        Example:
+            >>> provider = BraketProvider()
+            >>> backend = provider.get_backend("Aria 1", emulator=True)
+            >>> transpiled_circuit = transpile(circuit, backend=backend)
+            >>> backend.run(transpiled_circuit, shots=10).result().get_counts()
+            {"100": 6, "001": 4}
+
+        Args:
+            device (AwsDevice): The Braket device to emulate. Its properties are used
+                to build the Qiskit ``Target`` and the emulator backend.
+            provider: Qiskit provider for this backend. Default: ``None``.
+            name (str | None): Name of backend. Default: ``None``.
+            description (str | None): Description of backend. Default: ``None``.
+            online_date (datetime | None): Online date. Default: ``None``.
+            backend_version (str | None): Backend version. Default: ``None``.
+            **fields: Extra arguments.
+        """
+        emulator = device.emulator
+        # Emulators have no device properties, so the base initializer (which probes
+        # ``device.properties.action``) is bypassed. Emulators run locally and do not
+        # support program sets.
+        BackendV2.__init__(
+            self,
+            provider=provider,
+            name=name or device.name,
+            description=description or f"Emulator for AWS Device: {device.name}.",
+            online_date=online_date,
+            backend_version=backend_version,
+            **fields,
+        )
+        self._device = emulator
+        self._supports_program_sets = False
+        self._target = aws_device_to_target(device=device)
+        self._qubit_labels = (
+            tuple(sorted(device.topology_graph.nodes)) if device.topology_graph else None
+        )
+        self.status = emulator.status
+
+    @property
+    def emulator(self) -> bool:
+        """bool: Whether this backend executes on a device emulator. Always ``True``."""
+        return True
+
+    @property
+    def target(self) -> Target:
+        return self._target
+
+    @property
+    def max_circuits(self) -> None:
+        return None
+
+    @property
+    def qubit_labels(self) -> tuple[int, ...] | None:
+        return self._qubit_labels
+
+    @classmethod
+    def _default_options(cls) -> Options:
+        return Options()
+
+    @property
+    def dtm(self) -> float:
+        raise NotImplementedError(
+            f"System time resolution of output signals is not supported by {self.name}."
+        )
+
+    @property
+    def meas_map(self) -> list[list[int]]:
+        raise NotImplementedError(f"Measurement map is not supported by {self.name}.")
+
+    def qubit_properties(self, qubit: int | list[int]) -> QubitProperties | list[QubitProperties]:
+        raise NotImplementedError
+
+    def drive_channel(self, qubit: int):
+        raise NotImplementedError(f"Drive channel is not supported by {self.name}.")
+
+    def measure_channel(self, qubit: int):
+        raise NotImplementedError(f"Measure channel is not supported by {self.name}.")
+
+    def acquire_channel(self, qubit: int):
+        raise NotImplementedError(f"Acquire channel is not supported by {self.name}.")
+
+    def control_channel(self, qubits: Iterable[int]):
+        raise NotImplementedError(f"Control channel is not supported by {self.name}.")
+
+    def run(
+        self,
+        run_input: QuantumCircuit | list[QuantumCircuit],
+        *,
+        shots: int = 1024,
+        **options,
+    ) -> BraketQuantumTask:
+        """Execute ``QuantumCircuit``s on the device emulator.
+
+        Args:
+            run_input (QuantumCircuit | list[QuantumCircuit]): ``QuantumCircuit`` or list
+                of ``QuantumCircuit``s.
+            shots (int): Number of measurement repetitions. Default: ``1024``.
+            **options: Extra options passed to the emulator.
+
+        Returns:
+            BraketQuantumTask: The task tracking emulated execution.
+        """
+        convert_input = [run_input] if isinstance(run_input, QuantumCircuit) else list(run_input)
+        verbatim = options.pop("verbatim", False)
+        circuits: list[Circuit] = [
+            to_braket(circ, target=self._target if not verbatim else None, verbatim=verbatim)
+            for circ in convert_input
+        ]
+
+        if shots == 0:
+            circuits = [x.state_vector() for x in circuits]
+        if "meas_level" in options:
+            self._validate_meas_level(options["meas_level"])
+            del options["meas_level"]
+        tasks = []
+        try:
+            for circuit in circuits:
+                task: LocalQuantumTask = self._device.run(task_specification=circuit, shots=shots)
+                tasks.append(task)
+
+        except Exception as ex:
+            logger.error("During creation of tasks an error occurred: %s", ex)
+            logger.error("Cancelling all tasks %d!", len(tasks))
+            for task in tasks:
+                logger.error("Attempt to cancel %s...", task.id)
+                task.cancel()
+                logger.error("State of %s: %s.", task.id, task.state())
+            raise
+
+        task_id = _TASK_ID_DIVIDER.join(task.id for task in tasks)
+
+        return BraketQuantumTask(
+            task_id=task_id,
+            tasks=tasks,
+            backend=self,
+            shots=shots,
+        )
+
+
 class BraketAwsBackend(BraketBackend[AwsDevice]):
     """Runs quantum circuits on the Amazon Braket service."""
 
@@ -281,6 +444,26 @@ class BraketAwsBackend(BraketBackend[AwsDevice]):
             backend=self,
             tasks=[AwsQuantumTask(arn=task_id) for task_id in task_ids],
         )
+
+    @property
+    def emulator(self) -> BraketEmulatorBackend:
+        """Return a local emulator backend for this device.
+
+        The emulator mimics the gate set, connectivity and noise of the device while
+        executing circuits locally, providing a soft check that a program can run on the
+        target device without submitting tasks to the Amazon Braket service. Emulators
+        are only available for QPUs, not for Braket managed simulators.
+
+        Returns:
+            BraketEmulatorBackend: A backend that runs circuits on this device's emulator.
+
+        Example:
+            >>> backend = BraketProvider().get_backend("Aria 1")
+            >>> emulator_backend = backend.emulator
+            >>> emulator_backend.run(transpiled_circuit, shots=10).result().get_counts()
+            {"100": 6, "001": 4}
+        """
+        return BraketEmulatorBackend(device=self._device, provider=self.provider)
 
     @property
     def target(self) -> Target:
--- a/qiskit_braket_provider/providers/braket_backend.py
+++ b/qiskit_braket_provider/providers/braket_backend.py
@@ -52,6 +52,7 @@ class BraketBackend(BackendV2, ABC, Generic[T]):
     def __init__(self, device: T, name: str, **fields) -> None:
         super().__init__(name=name, **fields)
         self._device = device
+        self._qubit_labels: tuple[int, ...] | None = None
         self._supports_program_sets = (
             DeviceActionType.OPENQASM_PROGRAM_SET in self._device.properties.action
         )
@@ -66,7 +67,7 @@ class BraketBackend(BackendV2, ABC, Generic[T]):
 
         Unlike the qubits in the target, these labels are not necessarily contiguous.
         """
-        return None
+        return self._qubit_labels
 
     def _validate_meas_level(self, meas_level: enum.Enum | int) -> None:
         if isinstance(meas_level, enum.Enum):
@@ -105,9 +106,25 @@ class BraketBackend(BackendV2, ABC, Generic[T]):
 
 
 class BraketLocalBackend(BraketBackend[LocalSimulator]):
-    """Runs quantum circuits on the Braket local simulator."""
+    """Runs quantum circuits on a Braket local simulator or device emulator.
 
-    def __init__(self, name: str = "default", **fields) -> None:
+    By default the backend wraps a local :class:`~braket.devices.LocalSimulator`
+    selected by ``name``. Passing a pre-built ``device``, such as the emulator
+    returned by :meth:`~braket.aws.AwsDevice.emulator`, runs circuits on that device
+    instead. An emulator mimics the gate set, connectivity and noise of an
+    ``AwsDevice`` while executing locally; it exposes no device properties of its own,
+    so the ``target`` and ``qubit_labels`` of the source device are passed in.
+    """
+
+    def __init__(
+        self,
+        name: str = "default",
+        *,
+        device: Device | None = None,
+        target: Target | None = None,
+        qubit_labels: tuple[int, ...] | None = None,
+        **fields,
+    ) -> None:
         """Initialize the backend.
 
         Example:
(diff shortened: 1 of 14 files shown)


sesmart:
Hi @zkasuran , thank you for your PR. Please note that we are leaning towards a more minimal solution, which is not far from what you have :) and should reduce coverage a bit. Also I would simply try `tox`, to capture all of the scripts in pyproject.toml.

zkasuran:
Thanks @sesmart. I went minimal: the `shots=0` branch on the emulator was copied from `BraketLocalBackend` but the emulator can't return a state vector, so it failed at runtime. Keeping it raised `AttributeError: 'StateVector' object has no attribute 'target'` in the emulator's result-type validator, and dropping it instead hit `ValueError: No result types specified for circuit and shots=0`. So I replaced the branch with an explicit `QiskitBraketException` that points users to `BraketLocalBackend` for exact simulation.

Net diff is +31/-2 across the backend and one test file. I ran `tox`: the full suite is green (330 passed, 4 skipped), coverage is back to 100%, and `ruff format`/`ruff check` are clean.

AI disclosure: this change was implemented with help from Claude (Anthropic), which drafted the fix and tests. I verified it locally with `tox` (pytest + coverage 100% + ruff) before pushing.

zkasuran:
Went another round of minimal, since the first cut only removed the shots=0 branch. `BraketEmulatorBackend` now derives from `BraketLocalBackend` instead of duplicating its run path and stubs, which is also what point 3 of #325 asks for: the emulator type is a local backend. The class went from 192 to 122 lines and the PR from +516/-6 to +383/-7. Most of what remains is the mock device capabilities that a real `LocalEmulator` can be built from, so the run tests execute a real emulator rather than stubs.

The rewrite surfaced two real fixes. The SDK defines `AwsDevice.emulator()` as a method and the previous version read it as an attribute, which the fully mocked tests could not catch. And `backends(emulator=True)` now excludes managed simulators, since `device.emulator()` raises a `ValueError` for them.

`tox`: 325 passed, 4 skipped, coverage 100%, ruff clean. `tox -e mypy` also clean.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -208,6 +208,108 @@ def run(
         )
 
 
+class BraketEmulatorBackend(BraketLocalBackend):
```
sesmart: So right now we are not looking to have a new class - can we modify the LocalBackend in such a way that we can just pass the emulator as the device? There are a few subtitles, but it overall is not backwards imcompatible with a a new Emulator Backend class.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -282,6 +384,26 @@ def retrieve_job(self, task_id: str) -> BraketQuantumTask:
...
+        Returns:
+            BraketEmulatorBackend: A backend that runs circuits on this device's emulator.
+
+        Example:
+            >>> backend = BraketProvider().get_backend("Aria 1")
+            >>> emulator_backend = backend.emulator
+            >>> emulator_backend.run(transpiled_circuit, shots=10).result().get_counts()
+            {"100": 6, "001": 4}
+        """
+        return BraketEmulatorBackend(device=self._device, provider=self.provider)
```
sesmart: I don't think you need the provider here?

qiskit_braket_provider/providers/braket_provider.py
```diff
@@ -30,19 +30,23 @@ class BraketProvider:
...
             **kwargs: dict with additional options for filtering and storing aws session
         Returns:
-            BraketAwsBackend: a backend matching the filters.
+            BraketAwsBackend | BraketEmulatorBackend: a backend matching the filters.
         Raises:
             QiskitBackendNotFoundError: if no backend could be found or
             more than one backend matches the filters.
         """
-        backends = self.backends(name=name, **kwargs)
+        backends = self.backends(name=name, emulator=emulator, **kwargs)
```
sesmart: could you also just get the first backend and then use the defined method above? I think that follows more how we get emulators in the BDK

test/unit_tests/test_braket_provider.py
```diff
@@ -242,3 +246,59 @@ def test_kraus_target_simulator(self):
...
+        device.provider_name = "provider1"
+        device.properties = capabilities
+        device.gate_calibrations = None
+        device.type = "QPU"
+        device.topology_graph = mock_emulator_topology()
+        device.emulator.return_value = LocalEmulator.from_device_properties(capabilities)
+        return device
+
+    @patch("qiskit_braket_provider.providers.braket_provider.AwsDevice.get_devices")
+    def test_provider_backends_emulator(self, mock_get_devices: MagicMock):
```
sesmart: I don't think this test will be necessary if we don't define emulator as a kwarg in backends

sesmart (commented):
Thanks @zkasuran, sorry I was OOO the last two days. Please let me know if there are any questions.

zkasuran:
Thanks @sesmart, this all made sense. I reworked it along those lines.

- No new class. `BraketEmulatorBackend` is gone. `BraketLocalBackend` now takes an optional pre-built `device`, so you pass the emulator from `AwsDevice.emulator()` straight in, along with the `target` and `qubit_labels` of the source device (the emulator exposes no properties of its own to build those from). It still defaults to a `LocalSimulator` when no device is given. An `emulator` property reports which mode it is in. `shots=0` is rejected only in emulator mode, since the emulator reproduces QPU behavior and can't return a state vector.
- Dropped the `provider` from the emulator construction in `BraketAwsBackend.emulator`.
- `get_backend(name, emulator=True)` now grabs the single backend then returns `backend.emulator`, the same way you reach an emulator in the BDK. `backends()` no longer has an `emulator` kwarg.
- Removed `test_provider_backends_emulator` since that kwarg is gone.

One small thing the collapse let me tidy: `qubit_labels` was duplicated on both backends, so it now lives on the `BraketBackend` base.

`tox`: 324 passed, 4 skipped, coverage 100%. `tox -e mypy` and ruff are clean.

AI disclosure: I implemented this with help from Claude (Anthropic), which drafted the refactor and updated the tests. I verified it locally with `tox` (pytest + 100% coverage + ruff) and `tox -e mypy` before pushing.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -119,13 +136,35 @@ def __init__(self, name: str = "default", **fields) -> None:
...
-        self._gateset = self.get_gateset()
-        self.status = self._device.status
+        self._is_emulator = device is not None
+        if device is None:
+            device = LocalSimulator(backend=name)
+            target = target or local_simulator_to_target(device)
+        BackendV2.__init__(self, name=name or device.name, **fields)
+        self._device = device
+        self._supports_program_sets = (
+            not self._is_emulator
```
sesmart: Emulator's actually support program sets - there is a certain amount of payload validation (i.e. verbatim circuits + connectivity + executable & shot properties) that the emulator does. So program sets are definitely part of that.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -105,9 +106,25 @@ def _run_program_set(
...
 class BraketLocalBackend(BraketBackend[LocalSimulator]):
-    """Runs quantum circuits on the Braket local simulator."""
+    """Runs quantum circuits on a Braket local simulator or device emulator.
 
-    def __init__(self, name: str = "default", **fields) -> None:
+    By default the backend wraps a local :class:`~braket.devices.LocalSimulator`
+    selected by ``name``. Passing a pre-built ``device``, such as the emulator
+    returned by :meth:`~braket.aws.AwsDevice.emulator`, runs circuits on that device
+    instead. An emulator mimics the gate set, connectivity and noise of an
+    ``AwsDevice`` while executing locally; it exposes no device properties of its own,
```
sesmart: Don't need this last sentence here.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -119,13 +136,35 @@ def __init__(self, name: str = "default", **fields) -> None:
 
         Args:
             name (str): Name of backend. Default: ``default``.
+            device (Device | None): A pre-built local device to run on, such as the
```
sesmart: I'd modify these descriptions a bit. The functionality here is a bit more generic than only the emulator. You could specify a custom backend, qubit labels, etc., and it would work find. I would change references below from emulator to the device.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -119,13 +136,35 @@ def __init__(self, name: str = "default", **fields) -> None:
...
+            qubit_labels (tuple[int, ...] | None): Qubit labels of the emulated device,
+                in ascending order. Default: ``None``.
             **fields: Extra arguments.
         """
-        simulator = LocalSimulator(backend=name)
-        super().__init__(simulator, name or simulator.name, **fields)
-        self._target = local_simulator_to_target(self._device)
-        self._gateset = self.get_gateset()
-        self.status = self._device.status
+        self._is_emulator = device is not None
```
sesmart: Or just check instance of LocalEmulator / Emulator

qiskit_braket_provider/providers/braket_backend.py
sesmart: This part is the one I am the most on the fence about.....I think there should be a way to infer this. You could use .emulator to get to the emulator._device (or I forget exactly the property, but it does have a private reference to the actual device). There also is a program set validation pass somewhere as well.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -282,6 +328,32 @@ def retrieve_job(self, task_id: str) -> BraketQuantumTask:
             tasks=[AwsQuantumTask(arn=task_id) for task_id in task_ids],
         )
 
+    @property
```
sesmart: We can change this back to a method - so no property tag is okay.

qiskit_braket_provider/providers/braket_provider.py
```diff
@@ -47,7 +52,10 @@ def get_backend(self, name: str | None = None, **kwargs) -> BraketAwsBackend:
             raise QiskitBackendNotFoundError("More than one backend matches the criteria")
         if not backends:
             raise QiskitBackendNotFoundError("No backend matches the criteria")
-        return backends[0]
+        backend = backends[0]
+        if emulator:
+            return cast("BraketAwsBackend", backend).emulator
```
sesmart: Why do we need cast here? It's the local backend type, which is okay?

test/unit_tests/mocks.py
```diff
@@ -357,6 +357,96 @@
...
+            {"fidelityType": {"name": "READOUT"}, "fidelity": 0.95},
+        ],
+    }
+
+
+# A small, self-consistent set of capabilities (2 qubits, fully connected) that
+# satisfies the validation performed by ``LocalEmulator.from_device_properties``,
+# which a plain QPU mock does not (it requires ``standardized`` per-qubit metrics,
+# supported result types and qubit counts that all agree).
+MOCK_EMULATOR_CAPABILITIES_JSON = {
```
sesmart, line 371: For a little more validation I would change the qubit numbers to 1 & 2.

sesmart (commented):
Thanks @zkasuran , almost there. Have you tried running this locally for a few emulators?

zkasuran:
Thanks @sesmart, this was a good round. Reworked along your notes, with one finding worth flagging.

**Program sets (the part you were on the fence about).** You were right that the emulator does program-set validation, so hard-coding `not self._is_emulator` was wrong. The emulator carries no `properties`. In braket-sdk 1.119 it also keeps no reference back to the source `AwsDevice` (it only holds `_backend`, a `LocalSimulator`, plus its pass manager and noise model), so I could not reach the device through it. What it does carry is the program-set validation pass you mentioned: I read `ProgramSetValidator.device_actions` off the emulator and report support from there, so an emulator of a program-set QPU now works with `BraketSampler` and `BraketEstimator`. One catch: `Emulator.run` defaults `shots` to 0 and forwards it, where `LocalSimulator.run` defaults to `None` and reads the per-executable count off the program set. So the primitives now pass `shots=None` explicitly, which is a no-op for the local simulator and AWS paths and lets the emulator run the set. There is a test that samples a program set on an emulator end to end.

**Inferring target and qubit_labels.** Same root cause: the emulator exposes nothing to rebuild them from. They are already inferred for the normal path though, since `BraketAwsBackend.emulator()` builds the emulator from the source device and passes its target and qubit_labels in, so callers never supply them. They stay as optional args only for the direct `BraketLocalBackend(device=...)` case.

**Changing the mock to qubits 1 and 2 caught a real bug.** The local run path was not passing `qubit_labels` to `to_braket`, so circuits were emitted on qubits 0,1 and the emulator rejected them as off topology. Fixed by threading `qubit_labels` through, the same way the AWS path does it. Good call.

The rest:
- `emulator` is a method again on `BraketAwsBackend`, matching `AwsDevice.emulator()`.
- Dropped the cast in `get_backend`, narrowed with an isinstance assert since only QPUs have emulators.
- `_is_emulator` is now `isinstance(device, Emulator)`.
- Trimmed the class docstring and reworded the arg docs to talk about a device rather than an emulator.

`tox`: 326 passed, 4 skipped, coverage 100%. `tox -e mypy` and `tox -e linters` are clean.

AI disclosure: I implemented this with help from Claude (Anthropic), which drafted the changes and tests. I verified it locally with `tox` (pytest + 100% coverage + ruff) and `tox -e mypy` before pushing.

zkasuran:
Yes, ran it locally against a few emulators. No AWS needed, the emulator builds offline on a `LocalSimulator` backend.

- Bell state on a noisy 3-qubit emulator (gateset + qubit-count passes plus depolarizing noise on the density-matrix sim) through `BraketLocalBackend(device=emulator)`: counts come out `{'00': 90, '11': 109, '10': 1}`, the expected Bell distribution.
- Gateset validation fires: a `CY` is rejected with "Gate CY is not a supported gate for this device."
- Qubit-count validation fires: a 4-qubit circuit on a 3-qubit emulator is rejected with "Circuit must use at most 3 qubits."
- The primitives path is covered end to end in the suite: `test_emulator_backend_with_sampler` runs `BackendSamplerV2` on an emulator and `test_emulator_backend_with_braket_sampler` samples a program set on a program-set emulator. Full unit suite is 326 passed, 4 skipped locally.

Happy to fold the standalone script into the diff as an example if useful. Or leave it out, your call.

sesmart:
Interesting catch on the shots =0 element. I think this definitely would need to be passed through, and there is no harm as it simply is validated against shots_per_executable * num_executables.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -46,12 +48,27 @@
 T = TypeVar("T", bound=Device, covariant=True)  # noqa: PLC0105
 
 
+def _emulator_supports_program_sets(emulator: Emulator) -> bool:
+    """Whether the device emulated by ``emulator`` accepts OpenQASM program sets.
+
+    An emulator exposes no ``properties`` of its own. It mirrors the source device's
```
sesmart: Just a 1-sentence on the private method is fine.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -174,11 +228,23 @@ def run(
...
+                qubit_labels=self._qubit_labels,
+                verbatim=verbatim,
+            )
             for circ in convert_input
         ]
 
         if shots == 0:
+            if self._is_emulator:
+                raise QiskitBraketException(
+                    "The device emulator does not support shots=0. It reproduces QPU "
```
sesmart: This isn't quite true, but I think the fact that the emulator supports 0 shots might be a bug.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -282,6 +348,31 @@ def retrieve_job(self, task_id: str) -> BraketQuantumTask:
             tasks=[AwsQuantumTask(arn=task_id) for task_id in task_ids],
         )
 
+    def emulator(self) -> BraketLocalBackend:
+        """Return a local emulator backend for this device.
```
sesmart: Shorten

qiskit_braket_provider/providers/braket_sampler.py
```diff
@@ -101,8 +101,11 @@ def run(
             parameter_indices.append(indices)
         shots_per_executable = pub_shots if pub_shots is not None else shots
         program_set = ProgramSet(circuit_bindings, shots_per_executable=shots_per_executable)
+        # shots live on the program set; pass shots=None so every device (including the
```
sesmart: remove comments

test/unit_tests/mocks.py
```diff
@@ -357,6 +357,119 @@
...
+        "T1": {"value": 28.9, "unit": "us"},
+        "T2": {"value": 44.5, "unit": "us"},
+        "oneQubitFidelity": [
+            {"fidelityType": {"name": "RANDOMIZED_BENCHMARKING"}, "fidelity": 0.999},
+            {"fidelityType": {"name": "READOUT"}, "fidelity": 0.95},
+        ],
+    }
+
+
+# A small, self-consistent set of capabilities (2 qubits, fully connected) that
```
sesmart: remove comments

test/unit_tests/mocks.py
```diff
@@ -357,6 +357,119 @@
...
+})
+
+
+def mock_emulator_topology() -> DiGraph:
+    """Return the topology graph matching ``mock_emulator_capabilities``."""
+    graph = from_dict_of_lists({"1": [2], "2": [1]}, create_using=DiGraph())
+    return relabel_nodes(graph, {n: int(n) for n in graph.nodes})
+
+
+# Program-set action for a source device that supports OpenQASM program sets. The
```
sesmart: remove comments

test/unit_tests/mocks.py
```diff
@@ -357,6 +357,119 @@
...
+        "qubitCount": 2,
+        "nativeGateSet": ["rx", "rz", "cz"],
+        "connectivity": {
+            "fullyConnected": True,
+            "connectivityGraph": {"1": ["2"], "2": ["1"]},
+        },
+    },
+    "deviceParameters": {},
+}
+MOCK_EMULATOR_STANDARDIZED_PROPERTIES = StandardizedGateModelQpuDeviceProperties.parse_obj({
```
sesmart, line 407: space here?

qiskit_braket_provider/providers/braket_backend.py
sesmart: Also we can just shorten this.

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -46,12 +48,27 @@
...
+    )
+
+
 class BraketBackend(BackendV2, ABC, Generic[T]):
     """Base Qiskit backend for Amazon Braket devices."""
 
     def __init__(self, device: T, name: str, **fields) -> None:
         super().__init__(name=name, **fields)
         self._device = device
+        self._qubit_labels: tuple[int, ...] | None = None
```
sesmart, line 70: Is this and the below one necessary?

qiskit_braket_provider/providers/braket_backend.py
```diff
@@ -46,12 +48,27 @@
 T = TypeVar("T", bound=Device, covariant=True)  # noqa: PLC0105
 
 
+def _emulator_supports_program_sets(emulator: Emulator) -> bool:
```
sesmart, line 51: Okay, I don't like that we have to access private passes here. I might be inclined just to set it to true if there is an emulator and let the emulator do the validation.

sesmart: This also does solve the min version issue that you just went through, so, apologies.

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -115,8 +115,11 @@ def run(
 
         shots = int(np.ceil(1.0 / (pub_precision if pub_precision is not None else precision) ** 2))
         program_set = ProgramSet(all_bindings, shots_per_executable=shots)
+        # shots live on the program set; pass shots=None so every device (including the
```
sesmart: remove comments

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -115,8 +115,11 @@ def run(
...
+        # emulator, whose run() defaults to 0) takes its per-executable count from there.
+        run_options = {"shots": None, **self._options}
```
sesmart: I would just add shots to options, and similarly below.

zkasuran:
Thanks @sesmart. Took your four notes.

**shots=0 (`braket_backend.py`).** You were right that the message wasn't quite true, so I dropped it. shots=0 now passes straight through to the device run instead of being intercepted. Before leaning on the "no harm" read I ran it against amazon-braket-sdk 1.118 across the paths an emulator can take:

- plain circuit, shots=0 on the emulator's density-matrix backend: `ValueError: No result types specified for circuit and shots=0`
- a `StateVector` result type, shots=0: `TypeError: result type StateVector is not supported by DensityMatrixSimulator`
- as a program set, `shots_per_executable=0`: `ValueError: No per-executable shots defined`

So in 1.118 the emulator rejects 0 shots on every path rather than running it. That matches your "might be a bug" hunch from the other direction: there is no path where it silently returns a wrong result, it just errors. Passing through hands the user that native SDK error. The one piece I kept is that the `state_vector()` result type is added only for the plain local simulator, not the emulator, since its density-matrix backend can't produce one (the `TypeError` above). So the change is just `if shots == 0 and not self._is_emulator:`, and the shots=0 test now asserts the call is forwarded to the device with `shots=0` and no state-vector result type bolted on.

**`emulator()` docstring.** Shortened to two lines plus the example.

**sampler / estimator comments.** Removed in both places. Kept the `{"shots": None, **options}` merge so the emulator still reads its per-executable count off the program set.

tox: 326 passed, 4 skipped, coverage 100%. mypy and linters clean.

AI disclosure: I implemented this with help from Claude (Anthropic). I verified it locally with tox (pytest + 100% coverage + ruff) and mypy. I also ran the shots=0 SDK probe above by hand before pushing.

sesmart:
@zkasuran  - can you remove the comments in the testing files and simplify some of the doc strings more?

zkasuran:
Done, thanks @sesmart.

- Comments: dropped both explanatory blocks on the emulator mocks in `mocks.py` (the capabilities block and the program-set action block). The mock structure reads fine without them.
- Docstrings: trimmed the `BraketLocalBackend` class docstring, the `_emulator_supports_program_sets` helper and `_resolve_compilation_args` down to a line or two, and tightened the `__init__` arg docs.

`tox`: 331 passed, 4 skipped, coverage 100%. `tox -e mypy` clean. ruff check and format clean.

AI disclosure: I made these edits with help from Claude (Anthropic) and verified them locally with `tox` before pushing.

sesmart (approved):
Thanks @zkasuran - nice work on catching some of the tangential issues and tricky cases. Please leave a comment on the original issue.

PR #337, opened 2026-06-08
feature: group qubit-wise commuting observables in BraketEstimator

### LINKED ISSUE #255: Group mutually commuting observables in BraketEstimator

BraketEstimator currently runs at least one executable for each observable in an EstimatorPub; this can be made more efficient by grouping mutually-commuting observables and running a single executable for each group.

LakshikkaNathan:
Fixes #255.

This adds an `abelian_grouping` option to `BraketEstimator.run`, turned on by default so it matches how Qiskit's `BackendEstimatorV2` already works.

At the moment the estimator runs one executable per Pauli term. With grouping on, terms that commute qubit-wise are measured together. For each group I build a single covering Pauli by OR-ing the group's z and x bits, run that one measurement, and then rebuild every term's expectation value from the shared shots. The reconstruction handles terms that show up in more than one observable, identity terms, and coefficients, and it still works when broadcasting over observables and parameters. Setting `abelian_grouping=False` falls back to the original per-term path, so existing behaviour is unchanged.

## Testing

I ran the unit tests and the linters locally. The existing tests that assert the per-term execution shape are pinned to `abelian_grouping=False` so they keep checking the old behaviour, and I added tests for the grouped path: one group collapsing into a single measurement, a term shared between two observables, a parameter sweep, identity-only and identity-with-active cases, Y-basis and mixed-basis groups, multiple groups, a four-qubit case, and a 2D observable-by-parameter broadcast. Grouped values are compared against `BackendEstimatorV2`.

* `pytest test/unit_tests -q`
* `tox -e linters_check`

## AI usage disclosure

In accordance with the unitaryHACK AI policy, I used an AI assistant (Claude) as a co-pilot while building this change, including help with the grouping approach and the tests. I read, ran, and verified all of the code locally, the grouped results are checked against Qiskit's `BackendEstimatorV2` in the test suite, and I can explain every part of the implementation.

### DIFF
--- a/qiskit_braket_provider/providers/braket_estimator.py
+++ b/qiskit_braket_provider/providers/braket_estimator.py
@@ -6,11 +6,20 @@ from functools import reduce
 from typing import TypeAlias
 
 import numpy as np
-from qiskit.primitives import BaseEstimatorV2, DataBin, EstimatorPubLike, PrimitiveResult, PubResult
+from qiskit.primitives import (
+    BaseEstimatorV2,
+    BasePrimitiveJob,
+    DataBin,
+    EstimatorPubLike,
+    PrimitiveResult,
+    PubResult,
+)
 from qiskit.primitives.containers.bindings_array import BindingsArray
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
+from qiskit.providers import JobStatus
 from qiskit.quantum_info import Pauli, PauliList, SparsePauliOp
 
+from braket.circuits import Circuit
 from braket.circuits.observables import Sum, Z
 from braket.program_sets import CircuitBinding, ParameterSets, ProgramSet
 from braket.tasks import ProgramSetQuantumTaskResult
@@ -31,11 +40,31 @@ _ResultMap: TypeAlias = dict[int, list[_ResultMapEntry]]
 
 
 @dataclass
+class _RoutingTarget:
+    pauli: Pauli
+    coeff: complex
+    positions_and_params: list
+
+
+@dataclass
+class _IdentityRouting:
+    coeff: complex
+    positions_and_params: list
+
+
+@dataclass
+class _QWCBindingMetadata:
+    param_idx_map: dict
+    routing_targets: list[_RoutingTarget]
+
+
+@dataclass
 class _PubMetadata:
     num_bindings: int
     binding_to_result_map: _ResultMap
     sum_binding_indices: set[int]
-    qwc_metadata: dict
+    qwc_metadata: dict[int, _QWCBindingMetadata]
+    identity_routing: list[_IdentityRouting]
 
 
 @dataclass
@@ -44,6 +73,39 @@ class _JobMetadata:
     pub_metadata: list[_PubMetadata]
     precision: float
     shots: int
+
+
+class _ConstantResultTask(BasePrimitiveJob[PrimitiveResult[PubResult], JobStatus]):
+    """
+    Job returned when every observable is constant (identity only), so there is
+    nothing to measure and no Braket task is submitted. The result is computed
+    analytically up front, unlike the lazy ``BraketPrimitiveTask``.
+    """
+
+    def __init__(self, result: PrimitiveResult[PubResult]) -> None:
+        super().__init__(job_id="constant")
+        self._result = result
+
+    def result(self) -> PrimitiveResult[PubResult]:
+        return self._result
+
+    def status(self) -> JobStatus:
+        return JobStatus.DONE
+
+    def cancel(self) -> None:
+        pass
+
+    def done(self) -> bool:
+        return True
+
+    def running(self) -> bool:
+        return False
+
+    def cancelled(self) -> bool:
+        return False
+
+    def in_final_state(self) -> bool:
+        return True
 
 
 class BraketEstimator(BaseEstimatorV2):
@@ -91,7 +153,7 @@ class BraketEstimator(BaseEstimatorV2):
         *,
         precision: float = _DEFAULT_PRECISION,
         abelian_grouping: bool = True,
-    ) -> BraketPrimitiveTask:
+    ) -> BasePrimitiveJob:
         """
         Run estimator on the given pubs.
 
@@ -113,29 +175,24 @@ class BraketEstimator(BaseEstimatorV2):
         pub_metadata = []  # Track which bindings belong to which pub
 
         for pub in coerced_pubs:
-            bindings, binding_to_result_map, sum_binding_indices, qwc_metadata = (
-                self._translate_pub(pub, abelian_grouping=abelian_grouping)
-            )
+            bindings, pub_meta = self._translate_pub(pub, abelian_grouping=abelian_grouping)
             all_bindings.extend(bindings)
-            pub_metadata.append(
-                _PubMetadata(
-                    num_bindings=len(bindings),
-                    binding_to_result_map=binding_to_result_map,
-                    sum_binding_indices=sum_binding_indices,
-                    qwc_metadata=qwc_metadata,
-                )
-            )
+            pub_metadata.append(pub_meta)
 
         shots = int(np.ceil(1.0 / (pub_precision if pub_precision is not None else precision) ** 2))
+        job_metadata = _JobMetadata(
+            pubs=coerced_pubs, pub_metadata=pub_metadata, precision=precision, shots=shots
+        )
+
+        # Every observable is constant: nothing to measure, so skip the device entirely
+        # and reconstruct the (purely identity) expectation values analytically.
+        if not all_bindings:
+            return _ConstantResultTask(BraketEstimator._translate_result(None, job_metadata))
+
         program_set = ProgramSet(all_bindings, shots_per_executable=shots)
         return BraketPrimitiveTask(
             self._backend._device.run(program_set, **self._options),
-            lambda result: BraketEstimator._translate_result(
-                result,
-                _JobMetadata(
-                    pubs=coerced_pubs, pub_metadata=pub_metadata, precision=precision, shots=shots
-                ),
-            ),
+            lambda result: BraketEstimator._translate_result(result, job_metadata),
             program_set,
         )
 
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -201,49 +217,155 @@ def _translate_pub(
...
+                ])
                 if param_values.data
                 else None
             )
-            binding_idx = len(bindings)
-            monomials = []
-            for ok, observable in zip(matching_obs_keys, braket_observables, strict=True):
-                if isinstance(observable, Sum):
+
+            if abelian_grouping:
```
speller26, line 290: Extract this complicated branch to a dedicated method

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -201,49 +217,155 @@ def _translate_pub(
...
+                                    "positions_and_params": obs_groups[ok],
+                                })
+
+                    qwc_metadata[binding_idx] = {
+                        "param_idx_map": param_idx_map,
+                        "routing_targets": [],
+                        "identity_routing": id_routing,
+                    }
+
+            else:
```
speller26, line 364: Also extract this non-grouping path to a dedicated method.

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -134,8 +147,8 @@ def _pub_precision(pubs: list[EstimatorPub]) -> float:
         return next(iter(precision_values))
 
     def _translate_pub(
-        self, pub: EstimatorPub
-    ) -> tuple[list[CircuitBinding], _ResultMap, set[int]]:
+        self, pub: EstimatorPub, abelian_grouping: bool = True
```
speller26: No need for a default value here, since it's a private method.

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -201,49 +217,155 @@ def _translate_pub(
...
+                        for ok in matching_obs_keys:
+                            op_from_dict = SparsePauliOp.from_list(list(obs_keys[ok].items()))
+                            for p, coeff in zip(
+                                op_from_dict.paulis, op_from_dict.coeffs, strict=True
+                            ):
+                                if p in obs_group:
+                                    routing_targets.append({
+                                        "pauli": p,
+                                        "coeff": coeff,
+                                        "positions_and_params": obs_groups[ok],
```
speller26: Prefer creating a metadata class like `_PubMetadata` or `_JobMetadata`; raw `dict`s with string keys are flimsy

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -201,49 +217,155 @@ def _translate_pub(
...
+                                    })
+
+                        qwc_metadata[binding_idx] = {
+                            "param_idx_map": param_idx_map,
+                            "routing_targets": routing_targets,
```
speller26: Same here

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -303,20 +425,55 @@ def _translate_result(
...
+
+                            flat_idx = np.unravel_index(position, broadcast_shape)
+                            evs[flat_idx] += coeff.real * term_expectation
+
+                    if "identity_routing" in meta:
+                        for id_target in meta["identity_routing"]:
+                            coeff = id_target["coeff"]
+                            for position, _ in id_target["positions_and_params"]:
+                                flat_idx = np.unravel_index(position, broadcast_shape)
+                                evs[flat_idx] += coeff.real * 1.0
```
speller26: Nice!

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -201,49 +217,155 @@ def _translate_pub(
...
+                                op_from_dict.paulis, op_from_dict.coeffs, strict=True
+                            ):
+                                if p in identity_paulis:
+                                    id_routing.append({
+                                        "coeff": coeff,
+                                        "positions_and_params": obs_groups[ok],
+                                    })
+                        qwc_metadata[target_binding]["identity_routing"] = id_routing
+
+                elif identity_paulis:
```
speller26: Why add trivial terms to the circuit binding at all if we already know the expectation value is 1 (times the coefficient), which you already use in result construction?

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done. Extracted into `_grouped_bindings` method

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done. Extracted into `_per_term_bindings` method

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done. Replaced with the `_RoutingTarget` and `_QWCBindingMetadata` dataclasses.

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done. Replaced with the `_IdentityRouting` dataclass.

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done. Removed the default

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Thanks! This moved to a single pub-level pass (`for term in pub_meta.identity_routing`) when I dropped the per-binding identity handling, but the idea is unchanged.

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Good call, you're right that there's no reason to bind a trivial term. Dropped the dummy binding and now collect identity terms as pub-level metadata, applied directly in result construction instead of through a measurement.

One consequence worth flagging: a fully-constant pub then produces zero bindings, and Braket's `ProgramSet` rejects an empty program list, so I added a small guard in `run()` that returns the analytic result via a lightweight `BasePrimitiveJob` (`_ConstantResultTask`) without calling the device. Added `test_abelian_grouping_pure_identity` for it. If you'd rather have constant-only observables raise instead, happy to switch.

LakshikkaNathan:
Thanks for the review @speller26. Pushed changes addressing all the comments:

- Extracted the two branches into `_grouped_bindings` and `_per_term_bindings`.
- Made `abelian_grouping` a required argument on `_translate_pub`.
- Replaced the string-keyed metadata dicts with the `_RoutingTarget`, `_IdentityRouting`, and `_QWCBindingMetadata` dataclasses.
- Dropped the identity dummy binding. Constant terms are now pub-level metadata applied analytically, and a fully-constant pub skips the device via a small `_ConstantResultTask` guard, covered by `test_abelian_grouping_pure_identity`.

One open decision: constant-only observables currently return the analytic value, but happy to raise instead if you'd prefer. Ready for another look when you have a chance.

speller26:
Looks like test coverage is failing

LakshikkaNathan:
Thanks! Added the missing test coverage. Back to 100% locally.

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -42,6 +75,39 @@ class _JobMetadata:
     shots: int
 
 
+class _ConstantResultTask(BasePrimitiveJob[PrimitiveResult[PubResult], JobStatus]):
```
speller26, line 78: Nice edge case coverage

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -27,11 +39,32 @@
 _ResultMap: TypeAlias = dict[int, list[_ResultMapEntry]]
 
 
+@dataclass
+class _RoutingTarget:
+    pauli: Pauli
+    coeff: complex
```
speller26: Only real coefficients are allowed

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -303,20 +524,50 @@ def _translate_result(
...
+                            param_set_idx = param_idx_map[param_indices] if param_indices else 0
+                            measured_entry = program_result[param_set_idx]
+                            measurements = measured_entry.measurements
+
+                            term_expectation = expectation_from_measurements(
+                                measurements, measured_qubits, braket_z_obs, targets
+                            )
+
+                            flat_idx = np.unravel_index(position, broadcast_shape)
+                            evs[flat_idx] += coeff.real * term_expectation
```
speller26: Why is a single Pauli being used with multiple task results? A single task should be responsible for multiple different Paulis, but here, the Pauli and coefficient are being used for multiple tasks.

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Thanks!

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Good catch, that was backwards. Flipped it to measurement-first so each covering measurement is read once and reconstructs all the Paulis routed to it, instead of one Pauli reaching across results. Values were already right (the index is the parameter set, not the Pauli), but it reads much better now.

qiskit_braket_provider/providers/braket_estimator.py
LakshikkaNathan: Done, typed both as `float`.

speller26:
Hi @LakshikkaNathan, thank you for the contribution! Unfortunately, while the logic in your PR was correct, another PR was closer to what we were looking for. In particular, there opportunities here to remove nested loops to increase the efficiency of both producing the groups and accumulating results.

However, this was a great effort, and you should be proud of what you've accomplished!

PR #342, opened 2026-06-14
feat: add abelian grouping of commuting observables to BraketEstimator

### LINKED ISSUE #255: Group mutually commuting observables in BraketEstimator

rylanmalarchick:
*Issue #, if available:* Closes #255

*Description of changes:*

This is a unitaryHACK 2026 bounty submission for #255.

`BraketEstimator` currently submits at least one Braket executable for every
observable in an `EstimatorPub`, even when several observables share a
measurement basis. This adds an opt-in `abelian_grouping` option that groups
mutually qubit-wise-commuting Pauli terms across a pub's observables and submits
a single executable per group per parameter set, mirroring Qiskit's
`BackendEstimatorV2`.

How it works:
- Collect the unique Pauli terms across the pub's observables and partition them
  with `SparsePauliOp.group_commuting(qubit_wise=True)`.
- Emit one `CircuitBinding` per group, measured in that group's shared
  single-qubit basis.
- Reconstruct each observable's expectation from the group's raw measurements via
  bitstring parity on each term's support.

Two design choices worth calling out:
- The default is `False`, which preserves the current executable layout and
  leaves every existing test unchanged. Pass `abelian_grouping=True` to enable.
  `BackendEstimatorV2` enables the equivalent grouping by default, so I am happy
  to flip the default to match if you prefer.
- The flag lives on the constructor rather than `run()`, because
  `BaseEstimatorV2.run(pubs, *, precision)` has a fixed signature and
  `BackendEstimatorV2` carries the equivalent setting as an option. Easy to move
  to `run()` if you would rather follow the issue text literally.

On semantics: grouped observables share one executable's shots, so each
observable keeps the same per-observable precision while total shots drop, and
estimates within a group become correlated (they come from the same measurement
samples). This is the expected behavior of abelian grouping.

*Testing done:*

All offline against `BraketLocalBackend` (no AWS account or region required):
- `test_abelian_grouping_reduces_executables`: six observables collapse from six
  executables to three, and the values still match the reference.
- `test_abelian_grouping_matches_reference_no_parameters` and
  `test_abelian_grouping_matches_reference_parameterized`: grouped expectation
  values match `BackendEstimatorV2` on the local simulator, covering a multi-term
  sum, an identity term, parameters, and broadcasting.
- `test_abelian_grouping_parity_reconstruction_exact`: feeds a hand-built
  measurement array through the reconstruction and asserts exact values, pinning
  the parity math and qubit-to-column mapping without simulator noise.
- Full estimator suite passes (21 passed); the 17 pre-existing tests are
  unchanged. `ruff check` and `ruff format` are clean.

AI assistance: I used an AI coding assistant (Claude Code) while implementing and
testing this, with a human in the loop throughout. I reviewed every line, made
the design decisions above, and can explain and defend the implementation.
Disclosed per unitaryHACK's AI policy.

#### General

- [x] I have updated any necessary documentation (the new option is documented in the `BraketEstimator` docstring)

#### Tests

- [x] I have added tests that prove my feature works
- [x] I have checked that my tests are not configured for a specific region or account

### DIFF
--- a/qiskit_braket_provider/providers/braket_estimator.py
+++ b/qiskit_braket_provider/providers/braket_estimator.py
@@ -329,9 +329,7 @@ class BraketEstimator(BaseEstimatorV2):
             grouped_colmap (dict): Per-binding qubit-to-column maps to populate (mutated in place).
         """
         # Collect every unique Pauli label across observables that share these parameter sets.
-        unique_labels = list({
-            label for ok in matching_obs_keys for label, _ in obs_keys[ok].items()
-        })
+        unique_labels = list({label for ok in matching_obs_keys for label in obs_keys[ok]})
         groups = SparsePauliOp(unique_labels).group_commuting(qubit_wise=True)
 
         # One binding per qubit-wise-commuting group, measured in that group's shared basis.
@@ -339,12 +337,12 @@ class BraketEstimator(BaseEstimatorV2):
         for group in groups:
             basis_label = BraketEstimator._group_basis_label(group)
             representative = translate_sparse_pauli_op(SparsePauliOp(basis_label))
-            support = sorted(BraketEstimator._pauli_support(basis_label))
+            basis_support = sorted(BraketEstimator._pauli_support(basis_label))
             binding_idx = len(bindings)
             bindings.append(
                 CircuitBinding(circuit, input_sets=parameter_sets, observables=[representative])
             )
-            grouped_colmap[binding_idx] = {qubit: col for col, qubit in enumerate(support)}
+            grouped_colmap[binding_idx] = {qubit: col for col, qubit in enumerate(basis_support)}
             grouped_binding_map[binding_idx] = []
             for label in group.paulis.to_labels():
                 label_to_binding[label] = binding_idx
--- a/qiskit_braket_provider/providers/braket_estimator.py
+++ b/qiskit_braket_provider/providers/braket_estimator.py
@@ -466,14 +466,13 @@ class BraketEstimator(BaseEstimatorV2):
             for local_binding_idx in range(num_bindings):
                 program_result = task_result[binding_offset + local_binding_idx]
 
-                if local_binding_idx in grouped_binding_map:
+                grouped_result_map_entry = grouped_binding_map.get(local_binding_idx)
+                if grouped_result_map_entry is not None:
                     # Abelian-grouped binding: recover each term's expectation from raw measurements
                     # via the parity of the bitstring on the term's support, then accumulate the
                     # coefficient-weighted contribution to its observable.
                     col_map = grouped_colmap[local_binding_idx]
-                    for position, param_idx, coeff, support in grouped_binding_map[
-                        local_binding_idx
-                    ]:
+                    for position, param_idx, coeff, support in grouped_result_map_entry:
                         if support:
                             measurements = program_result.entries[param_idx].measurements
                             cols = [col_map[qubit] for qubit in support]
--- a/qiskit_braket_provider/providers/braket_estimator.py
+++ b/qiskit_braket_provider/providers/braket_estimator.py
@@ -9,6 +9,7 @@ from qiskit.primitives.containers.bindings_array import BindingsArray
 from qiskit.primitives.containers.estimator_pub import EstimatorPub
 from qiskit.quantum_info import SparsePauliOp
 
+from braket.circuits import Circuit
 from braket.circuits.observables import Sum
 from braket.program_sets import CircuitBinding, ParameterSets, ProgramSet
 from braket.tasks import ProgramSetQuantumTaskResult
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -303,12 +461,32 @@ def _translate_result(
...
+                    # via the parity of the bitstring on the term's support, then accumulate the
+                    # coefficient-weighted contribution to its observable.
+                    col_map = grouped_colmap[local_binding_idx]
+                    for position, param_idx, coeff, support in grouped_binding_map[
+                        local_binding_idx
+                    ]:
+                        if support:
+                            measurements = program_result.entries[param_idx].measurements
+                            cols = [col_map[qubit] for qubit in support]
+                            parity = measurements[:, cols].sum(axis=1) % 2
```
speller26, line 480: Love this

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -303,12 +459,32 @@ def _translate_result(
...
-                num_observables = len(program_result.observables)
 
+                if local_binding_idx in grouped_binding_map:
+                    # Abelian-grouped binding: recover each term's expectation from raw measurements
+                    # via the parity of the bitstring on the term's support, then accumulate the
+                    # coefficient-weighted contribution to its observable.
+                    col_map = grouped_colmap[local_binding_idx]
+                    for position, param_idx, coeff, support in grouped_binding_map[
+                        local_binding_idx
+                    ]:
```
speller26:
```suggestion
                if local_binding_idx, grouped_result_map_entry in grouped_binding_map.items():
                    # Abelian-grouped binding: recover each term's expectation from raw measurements
                    # via the parity of the bitstring on the term's support, then accumulate the
                    # coefficient-weighted contribution to its observable.
                    col_map = grouped_colmap[local_binding_idx]
                    for position, param_idx, coeff, support in grouped_result_map_entry:
```

qiskit_braket_provider/providers/braket_estimator.py
```diff
@@ -26,12 +27,20 @@
 _ResultMapEntry: TypeAlias = tuple[int, int | None, int]
 _ResultMap: TypeAlias = dict[int, list[_ResultMapEntry]]
 
+# Reconstruction data for an abelian-grouped binding:
+# binding_index -> [(broadcast_position, parameter_set_index, coefficient, pauli_support)]
+# where pauli_support is the tuple of qubits the Pauli term acts on non-trivially.
+_GroupedResultMapEntry: TypeAlias = tuple[int, int, float, tuple[int, ...]]
+_GroupedResultMap: TypeAlias = dict[int, list[_GroupedResultMapEntry]]
```
speller26, line 34: Don't think this alias is necessary; `dict[int, list[_GroupedResultMapEntry]]` is already descriptive.

qiskit_braket_provider/providers/braket_estimator.py
rylanmalarchick: Good catch on the double lookup. I went with `grouped_result_map_entry = grouped_binding_map.get(local_binding_idx)` and `if grouped_result_map_entry is not None:` rather than iterating `.items()` directly: this loop also handles the ungrouped bindings by index and shares the `program_result = task_result[binding_offset + local_binding_idx]` fetch, so staying on the single `range(num_bindings)` pass avoids duplicating that fetch into a second loop. Kept `is not None` (rather than a truthiness check) so an empty entry list still takes the grouped branch, matching the original `in` semantics. Done in the latest push.

qiskit_braket_provider/providers/braket_estimator.py
rylanmalarchick: I'd lean toward keeping this alias. It mirrors the existing `_ResultMap` right above it (line 28, same `dict[int, list[...]]` shape), and the package already uses this composite-alias-from-an-entry-alias pattern elsewhere — e.g. `adapter.py`'s `_ParameterRestrictions: TypeAlias = dict[str, dict[_ParamKey, _QubitSet]]`. It also gives the `# binding_index -> [(...)]` structure comment a home and keeps the `_translate_pub` return signature (which already carries `_ResultMap`) symmetric. Happy to inline it if you'd rather — though for consistency that'd really mean inlining `_ResultMap` too, so I wanted to check your preference first.

speller26 (approved):
You'll need to run fix the formatting as well as

rylanmalarchick:
Thanks. On review, looks like your note got cut off after "fix the formatting as well as". The check-code-format job (ruff format --check + ruff check) is currently passing, so could you let me know what specifically needs reformatting?

speller26:
Ignore that; looks like a straggler comment that I was going to post before but only just made it through

PR #346, opened 2026-06-16
infra: restructure adapter.py into smaller files

aniksd-braket:
## Restructure adapter.py into smaller, focused modules
  
  Breaks up the 2,158-line `adapter.py` monolith into three new modules while maintaining full backward compatibility:
  
  
  | Module | Lines | Contents |
  |--------|-------|----------|
  | `gate_mappings.py` | 186 | Bidirectional gate lookup tables (`_BRAKET_TO_QISKIT_NAMES`, `_QISKIT_GATE_NAME_TO_BRAKET_GATE`, `_BRAKET_GATE_NAME_TO_QISKIT_GATE`, `_BRAKET_VERBATIM_BOX_NAME`) |
  | `qasm_context.py` | 558 | `QiskitProgramContext` — OpenQASM 3 → Qiskit circuit interpreter (conditions, loops, verbatim markers, mid-circuit measurement), `_sympy_to_qiskit` |
  | `target.py` | 513 | Device target construction (`aws_device_to_target`, `local_simulator_to_target`, `SubstitutedTarget`, gate connectivity/restrictions helpers) |
  | `compilation.py` | 373 | Compilation pipeline (`compile_circuits`, `_extract_verbatim_boxes`, `_restore_verbatim_boxes`, `_CompilationContext`, `_default_target`, validation helpers). |
  | `adapter.py` | 660 | Translation layer (`to_braket`, `to_qiskit`, `_translate_to_braket`), observable helpers, re-exports for backward compatibility |
  
  **Before:** `adapter.py` — 2,158 lines (single monolith)  
  **After:** 5 focused modules — 2,290 lines total (split + minor additions for imports/docstrings)

  **Functional code change** `_compile()` now take a Qiskit `QuantumCircuit` as an input argument and not a Braket `Circuit`

  ### What was tested
  
  - All 317 unit tests pass
  - Ruff linter clean
  - mypy shows only a pre-existing `arg-type` issue (unchanged from `main`)

## Merge Checklist

#### General

#### Tests


qiskit_braket_provider/providers/target.py
```diff
@@ -0,0 +1,505 @@
...
+    max_control = max_qubits if max_qubits is not None else inf
+    return {
+        controlled_gate
+        for control_count, gate_map in _CONTROLLED_GATES_BY_QUBIT_COUNT.items()
+        for controlled_gate, base_gate in gate_map.items()
+        if control_count <= max_control and base_gate in base_gateset
+    }
+
+
+# TODO: move target construction to a dedicated file; AwsDevice target construction is getting big
```
sesmart: You can remove this ^-^

test/unit_tests/test_adapter.py
```diff
@@ -961,21 +961,21 @@ def test_invalid_ctrl_state(self, mock_transpile: MagicMock):
         with pytest.raises(ValueError):
             to_braket(qiskit_circuit)
 
-    def test_get_controlled_gateset(self):
+    def testget_controlled_gateset(self):
```
nilsquet:
```suggestion
    def test_get_controlled_gateset(self):
```

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -0,0 +1,549 @@
...
+from sympy import Add, Expr, Mul, Pow, Symbol
+
+from braket.default_simulator.openqasm._helpers.arrays import convert_range_def_to_range
+from braket.default_simulator.openqasm._helpers.casting import cast_to
+from braket.default_simulator.openqasm._helpers.functions import (
+    evaluate_binary_expression,
+    evaluate_unary_expression,
+)
+from braket.default_simulator.openqasm.interpreter import (  # noqa: F401
+    Interpreter,
```
nilsquet: The `Interpreter` is not used in this file.

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -0,0 +1,229 @@
+"""Shared constants and gate mapping tables.
```
nilsquet, line 1: Consider renaming this module to gate_mappings.py. The contents are predominantly bidirectional gate lookup tables (mutable Qiskit instances and lambdas), not constants. Several genuine constants here (`_PAULI_MAP`, `_BRAKET_SUPPORTED_NOISES`, `_SYMPY_FUNCTION_TO_QISKIT_METHOD`, `_TRANSPILER_GATE_SUBSTITUTES`) have a single consumer each and would be cleaner placed in that consumer's module.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -1479,7 +320,7 @@ def _compile(
                 routing_method=effective_routing_method,
                 seed_transpiler=seed_transpiler,
             )
-    if isinstance(target, _SubstitutedTarget):
+    if isinstance(target, SubstitutedTarget):
```
nilsquet: To avoid scope creep, should we keep the private classes for now?

qiskit_braket_provider/providers/adapter.py
nilsquet: Please extract the compilation pipeline into `compilation.py` in this PR. Move `_compile`, `_extract_verbatim_boxes`,  `_restore_verbatim_boxes`, `_CompilationContext`, `_get_circuits`, `_check_positional`, `_validate_arguments`, `_default_target`, plus `_Translatable` and `_T`.  Drops `adapter.py` from ~1000 to ~650 lines.

nilsquet (changes requested):

qiskit_braket_provider/providers/adapter.py
aniksd-braket: I prefer changing this name. Keeping it private when it is clearly being imported doesn't make sense to me

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,373 @@
...
+        else:
+            qubit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.qubits]
+            clbit_indices = [transpiled_circuit.find_bit(q).index for q in instruction.clbits]
+            reconstructed_circuit.append(operation, qubit_indices, clbit_indices)
+
+    remaining_boxes = list(verbatim_box_iter)
+    if remaining_boxes:
+        raise ValueError(
+            f"Compiler error while processing verbatim boxes. Expected {barrier_count} "
+            "verbatim boxes, but found {len(verbatim_boxes)}."
```
nilsquet:
```suggestion
            f"verbatim boxes, but found {len(verbatim_boxes)}."
```

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,373 @@
...
+    add_measurements: bool = True,
+    circuit: _Translatable | Iterable[_Translatable] | None = None,
+    connectivity: list[list[int]] | None = None,
+    verbatim_box_name: str = _BRAKET_VERBATIM_BOX_NAME,
+    layout_method: str | None = None,
+    routing_method: str | None = None,
+    seed_transpiler: int | None = None,
+) -> _CompilationContext:
+    # Import here to avoid circular dependency
+    from qiskit_braket_provider.providers.adapter import to_qiskit
```
nilsquet: I'd narrow `compile_circuits` to take only `QuantumCircuit | Iterable[QuantumCircuit]` and let `to_braket` handle the `Circuit | Program | str` → `QuantumCircuit` conversion before calling it. The three tests passing QASM directly would become `compile_circuits(to_qiskit(qasm), target=...)`.
  
  `_compile` was private before this PR, so narrowing the new public API before it ships isn't breaking anyone. The payoff is that `compilation.py` no longer imports from `adapter.py` at all, and the split between the two modules actually means something.
  
  I tried this locally — all 331 tests pass, ruff is clean, and the net diff is +12 lines.

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -0,0 +1,186 @@
...
+
+This module contains the gate mapping dictionaries and constants used across
+adapter.py, target.py, and qasm_context.py.
```
nilsquet: The docstring lists adapter.py, target.py, and qasm_context.py as the consumers, but compilation.py also imports from this module now. Either add it or remove the explicit callout for all modules.

test/unit_tests/test_adapter.py
```diff
@@ -25,7 +25,7 @@
 
 from braket.aws import AwsDevice
 from braket.circuits import Circuit, Gate, GateCalibrations, Instruction
-from braket.circuits import compiler_directives as braket_compiler_directives
+from braket.circuits import compiler_directives as compile_circuitsr_directives
```
nilsquet: Is this change intentional?

test/unit_tests/test_adapter.py
```diff
@@ -961,7 +961,7 @@ def test_invalid_ctrl_state(self, mock_transpile: MagicMock):
         with pytest.raises(ValueError):
             to_braket(qiskit_circuit)
 
-    def test_get_controlled_gateset(self):
+    def test__get_controlled_gateset(self):
```
nilsquet:
```suggestion
    def test_get_controlled_gateset(self):
```

test/unit_tests/test_adapter.py
```diff
@@ -1841,7 +1841,7 @@ def test_empty_circuit_with_barrier(self):
             to_braket(circuit)
         self.assertIn("Cannot add global barrier to empty circuit", str(context.exception))
 
-    def test_compile_preserves_layout_with_verbatim_boxes(self):
+    def testcompile_circuits_preserves_layout_with_verbatim_boxes(self):
```
nilsquet:
```suggestion
    def test_compile_circuits_preserves_layout_with_verbatim_boxes(self):
```

test/unit_tests/test_adapter_dynamic_circuits.py
```diff
@@ -502,17 +502,19 @@ def mcm_target() -> Target:
         ),
     ],
 )
-def test_compile_preserves_if_else_ops(qasm: str, expected_if_else_count: int, mcm_target: Target):
-    """IfElseOps should survive compilation through _compile."""
-    result = _compile(qasm, target=mcm_target)
+def testcompile_circuits_preserves_if_else_ops(
```
nilsquet:
```suggestion
def test_compile_circuits_preserves_if_else_ops(
```

test/unit_tests/test_adapter_dynamic_circuits.py
```diff
@@ -502,17 +502,19 @@ def mcm_target() -> Target:
...
+    result = compile_circuits(qasm, target=mcm_target)
     compiled_circuit = result.circuits[0]
     if_else_ops = [
         instr for instr in compiled_circuit.data if isinstance(instr.operation, IfElseOp)
     ]
     assert len(if_else_ops) == expected_if_else_count
 
 
-def test_compile_if_else_branch_bodies_intact(mcm_target: Target):
+def testcompile_circuits_if_else_branch_bodies_intact(mcm_target: Target):
```
nilsquet:
```suggestion
def test_compile_circuits_if_else_branch_bodies_intact(mcm_target: Target):
```

test/unit_tests/test_adapter_dynamic_circuits.py
```diff
@@ -536,7 +538,7 @@ def test_compile_if_else_branch_bodies_intact(mcm_target: Target):
     assert _get_ops_with_qubits(false_body) == [("x", [1])]
 
 
-def test_compile_if_else_condition_preserved(mcm_target: Target):
+def testcompile_circuits_if_else_condition_preserved(mcm_target: Target):
```
nilsquet:
```suggestion
def test_compile_circuits_if_else_condition_preserved(mcm_target: Target):
```

nilsquet (changes requested):
The split itself is clean and behavior-preserving. Two things blocking:
  
  1. Test-file fallout from what looks like a global `_compile` → `compile_circuits` rename without word boundaries. It mangled an import alias and several test
  function names.
  
  2. The lazy import in `compile_circuits` is hiding a circular dependency that didn't exist before this PR. Worth fixing now while the boundary is fresh —
  narrowing the input type fixes it cleanly (details in the inline comment).
  
  One question: `QiskitProgramContext`, `SubstitutedTarget`, and `compile_circuits` are now public per the description. Is that intentional?

Otherwise nice cleanup — adapter.py was overdue for this.

sesmart:
Sorry, to @nilsquet point - do we need those functions to be public?

aniksd-braket:
@sesmart The compilation one yes, the others do not need to be but I strongly feel that they should be. The AbstractProgramContext and ProgramContext used in the default_simulator are both public. So from a consistency standpoint, it makes sense for the QiskitProgramContext to be public too. Similarly the SubstitutedTarget is used across modules, and I don't see why that shouldn't be public. It could also be used by customers if they want to update the target object for compilation

test/unit_tests/test_adapter_dynamic_circuits.py
```diff
@@ -536,7 +538,7 @@ def test_compile_if_else_branch_bodies_intact(mcm_target: Target):
     assert _get_ops_with_qubits(false_body) == [("x", [1])]
 
 
-def test_compile_if_else_condition_preserved(mcm_target: Target):
+def test_circuits_if_else_condition_preserved(mcm_target: Target):
```
nilsquet:
```suggestion
def test_compile_circuits_if_else_condition_preserved(mcm_target: Target):
```

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,340 @@
...
+    for braket_name, instruction in _BRAKET_GATE_NAME_TO_QISKIT_GATE.items():
+        if name := _BRAKET_TO_QISKIT_NAMES.get(braket_name.lower()):
+            target.add_instruction(instruction, name=name)
+    target.add_instruction(Measure())
+    target.add_instruction(Barrier(1))
+    return target
+
+
+def compile_circuits(
+    circuits: QuantumCircuit | Iterable[QuantumCircuit] = None,
```
nilsquet: `None` is no longer a matching type.

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,340 @@
...
+    target = Target(num_qubits=num_qubits)
+    for braket_name, instruction in _BRAKET_GATE_NAME_TO_QISKIT_GATE.items():
+        if name := _BRAKET_TO_QISKIT_NAMES.get(braket_name.lower()):
+            target.add_instruction(instruction, name=name)
+    target.add_instruction(Measure())
+    target.add_instruction(Barrier(1))
+    return target
+
+
+def compile_circuits(
```
nilsquet: If this is a public function, we need a docstring.

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,340 @@
...
+
+    if has_verbatim_boxes:
+        circuits = [
+            _restore_verbatim_boxes(circ, verbatim_boxes, verbatim_box_name)
+            if len(verbatim_boxes) > 0
+            else circ
+            for circ, verbatim_boxes in zip(circuits, all_verbatim_boxes, strict=False)
+        ]
+
+    return _CompilationContext(
```
nilsquet, line 320: If this is a public functions, we shouldn't return a private class instantiation.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -288,1218 +83,42 @@
...
-        verbatim_boxes: List of (box_circuit, original_qubit_indices) tuples
-        verbatim_box_name: The label name used to identify verbatim barriers
-
-    Returns:
-        Circuit with verbatim box contents restored
+def _get_circuits(
+    circuits: _Translatable | Iterable[_Translatable] | None,
+    circuit: _Translatable | Iterable[_Translatable] | None,
+    add_measurements: bool,
+    to_qiskit_fn: Callable,
```
nilsquet: Since `to_qiskit` is in the same file, we don't need to have it as a function parameter.

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -0,0 +1,558 @@
+"""OpenQASM 3 to Qiskit circuit context.
+
+This module provides _QiskitProgramContext which interprets OpenQASM 3 programs
```
nilsquet, line 3:
```suggestion
This module provides QiskitProgramContext which interprets OpenQASM 3 programs
```

qiskit_braket_provider/providers/target.py
```diff
@@ -0,0 +1,513 @@
...
+    ("rx", (pi / 2,)): qiskit_gates.SXGate(),
+    ("rx", (-pi / 2,)): qiskit_gates.SXdgGate(),
+}
+
+_ParamKey: TypeAlias = tuple[float | str, ...]
+_QubitSet: TypeAlias = set[tuple[int, ...]]
+_ParameterRestrictions: TypeAlias = dict[str, dict[_ParamKey, _QubitSet]]
+
+
+class SubstitutedTarget(Target):
```
nilsquet: If we agree to have this as a public class, we should add a docstring.

qiskit_braket_provider/providers/compilation.py
```diff
@@ -0,0 +1,362 @@
...
+    target = Target(num_qubits=num_qubits)
+    for braket_name, instruction in _BRAKET_GATE_NAME_TO_QISKIT_GATE.items():
+        if name := _BRAKET_TO_QISKIT_NAMES.get(braket_name.lower()):
+            target.add_instruction(instruction, name=name)
+    target.add_instruction(Measure())
+    target.add_instruction(Barrier(1))
+    return target
+
+
+def compile_circuits(
```
sesmart: Somehow I feel like `compile_circuits` not supporting `braket.Circuits` is not ideal. But at the same time this should not be a widely used function versus the qasm / braket endpoints, so `compile_circuits` is probably okay. 

I'd vote for `compile_`, but won't block.

yitchen-tim: +1 to making it private (`_compile` is a better convention that `compile_`) for two reasons:
1. The function signature is way too complicated to use. For a public API, I think we want a more specific one so that the function signature can be simpler. For example, we wrap the _compile function with compile to Braket or compile to Qiskit etc. In short, I expect that a public compile API will have a different function signature, and we don't need to rush to publish the API.
2. Public compile API should be reserved for top level feature. This function is an intermediate step and outputs CompilationContext which is not immediately usable.

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -0,0 +1,186 @@
...
+from qiskit.circuit import Parameter
+from qiskit.circuit.library import get_standard_gate_name_mapping
+from qiskit_ionq import ionq_gates
+
+from braket import experimental_capabilities as braket_expcaps
+from braket.circuits import gates as braket_gates
+from braket.circuits import noises as braket_noises
+from qiskit_braket_provider.providers import braket_instructions
+
+_BRAKET_TO_QISKIT_NAMES = {
```
laurencap, line 24: Very well scoped file. Nice to have this out of the way!

test/unit_tests/test_adapter.py
```diff
@@ -25,7 +25,7 @@
 
 from braket.aws import AwsDevice
```
laurencap, line 26: Nice to see that the test changes are very minor.

qiskit_braket_provider/providers/compilation.py
laurencap: +1. It was also previously called `_compile`, based on the tests, so even if we did want to change the name of this function, it should be done in a separate PR. When possible, a restructuring PR like this should have no changes beyond moving the files.

This is in part for ease of review (good job Scott for noticing this change!) and second for the way it affects git history. Big moves like this reset the git history of these lines. By keeping the code itself the same in this commit, it's easier to track functional changes like this before or after this PR, if needed.

qiskit_braket_provider/providers/target.py
```diff
@@ -0,0 +1,521 @@
...
+from qiskit_braket_provider.providers.gate_mappings import (
+    _BRAKET_GATE_NAME_TO_QISKIT_GATE,
+    _BRAKET_TO_QISKIT_NAMES,
+    _CONTROLLED_GATES_BY_QUBIT_COUNT,
+    _STANDARD_GATE_NAME_MAPPING,
+)
+
+_ADDITIONAL_U_GATES = {"u1", "u2", "u3"}
+
+_TRANSPILER_GATE_SUBSTITUTES: dict[tuple[str, tuple[float | str, ...]], Gate] = {
```
laurencap: Would this be appropriate for `gate_mappings`?

qiskit_braket_provider/providers/compilation.py
laurencap: On that note, were there other changes made to this CR beyond structure?

laurencap (changes requested):
Please move any code changes (besides moving code to different files / updating imports) to a new PR

qiskit_braket_provider/providers/compilation.py
aniksd-braket: Just one renaming of the QiskitProgramContext to be public. Can rollback and make that a separate PR

test/unit_tests/test_adapter_dynamic_circuits.py
```diff
@@ -504,7 +504,7 @@ def mcm_target() -> Target:
 )
 def test_compile_preserves_if_else_ops(qasm: str, expected_if_else_count: int, mcm_target: Target):
     """IfElseOps should survive compilation through _compile."""
-    result = _compile(qasm, target=mcm_target)
+    result = _compile(to_qiskit(qasm), target=mcm_target)
```
laurencap, line 507: This is another code change to how `_compile` works?

aniksd-braket: Yes it is. But it is an artifact of the move, required to avoid a circular dependency.

laurencap (approved):

nilsquet (approved):

PR #351, opened 2026-06-23
fix: Share parent bits in control-flow body circuits in to_qiskit

nilsquet:
*Description of changes:*
  
  Control-flow body sub-circuits used by `IfElseOp` / `ForLoopOp` / `WhileLoopOp` were being built with `QuantumCircuit(num_qubits, num_clbits)`, which allocates fresh register-backed `Qubit`/`Clbit` objects. Qiskit's
  drawer (and any tooling that recurses into `.blocks`) resolves inner bits in the parent's scope via object identity, so this raised `CircuitError: Could not locate provided bit` and made the circuit unprintable.
  
  Build each body with `copy_empty_like()` on the active circuit, then adopt any bits the body grew back into the parent. For `IfElseOp`, sync both bodies through the parent so they share the same bit layout. Two helpers on `_QiskitProgramContext` factor the pattern across the three call sites.
  
  *Testing done:*
  
  New parametrized test `test_control_flow_body_circuit_is_drawable` with three cases (if/else, while, for). Each asserts `print(qc)` works and that the control-flow body blocks reference the parent's exact bit objects. All three cases fail on mainline with the original `CircuitError`. `tox -e py313,linters_check` passes.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/qasm_context.py
+++ b/qiskit_braket_provider/providers/qasm_context.py
@@ -133,6 +133,17 @@ class _QiskitProgramContext(AbstractProgramContext):
             )
         return self._circuit_stack[0]
 
+    def _push_scoped_circuit(self) -> QuantumCircuit:
+        """Push an empty body circuit onto the stack that shares the parent's bit objects."""
+        body = self._active_circuit.copy_empty_like()
+        self._circuit_stack.append(body)
+        return body
+
+    def _extend_bits(self, target: QuantumCircuit, source: QuantumCircuit) -> None:
+        """Add to target any bits that source has beyond target's current bit list."""
+        target.add_bits(source.qubits[target.num_qubits :])
+        target.add_bits(source.clbits[target.num_clbits :])
+
     def add_qubits(self, name: str, num_qubits: int | None = 1) -> None:
         super().add_qubits(name, num_qubits)
         self._active_circuit.add_register(num_qubits)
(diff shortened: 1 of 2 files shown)


qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -322,15 +333,11 @@ def evaluate_condition(self, condition: Expression) -> Iterator[bool]:
                 f"Only Identifier, IndexExpression, and BinaryExpression (==) are supported."
             )
 
-        true_body = QuantumCircuit(main.num_qubits, main.num_clbits)
-        self._circuit_stack.append(true_body)
+        true_body = self._push_scoped_circuit()
```
laurencap, line 336: These changes not only fix a bug, but simplify the main logic at the same time. Nice job

laurencap (dismissed):

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -133,6 +133,17 @@ def circuit(self) -> QuantumCircuit:
             )
         return self._circuit_stack[0]
 
+    def _push_scoped_circuit(self) -> QuantumCircuit:
+        """Push an empty body circuit onto the stack that shares the parent's bit objects."""
+        body = self._active_circuit.copy_empty_like()
```
aniksd-braket, line 138: have you ensured this is not a shallow copy that could have unintended side effects?

nilsquet: Checked — copy_empty_like() makes the body refer to the same bits and registers as the parent and nothing else leaks: body-side append, add_bits, add_register, and global_phase don't touch the parent (and vice versa). 

@aniksd-braket — any specific side effect you had in mind? Happy to add a targeted test.

aniksd-braket: I was primarily concerned about the phase, could we add a test to ensure the phase isn't changed

nilsquet: I added a test that showcases the behavior.

aniksd-braket (approved):

PR #352, opened 2026-06-23
fix: Preserve if/measure inside verbatim box in to_qiskit

nilsquet:
*Description of changes:*
  
  `measure` and `if` inside `#pragma braket verbatim box { ... }` weren't being placed in the box body — they landed on the top-level circuit. Push a scoped body at `START_VERBATIM` so every construct appends inside the box; pop and wrap as a `BoxOp` at `END_VERBATIM`. `_resolve_condition_*` now reads `_active_circuit.clbits`.
  
Example:
```python
from qiskit_braket_provider.providers.adapter import to_qiskit
qasm = """
OPENQASM 3.0;
bit b;
#pragma braket verbatim
box {
    h $0;
    b = measure $0;
    if (b == 1) { x $0; }
}
"""
qc = to_qiskit(qasm)
```
  
  *Testing done:*
  
  Structural test (verbatim+if+measure preserved as nested `IfElseOp`) + round-trip test (QASM → Qiskit → QASM with non-identity clbit mapping survives). Both fail on mainline; `tox` passes here.
  
  **Known limitation:** `to_braket` of a verbatim BoxOp containing `if` raises `NotImplementedError` — Braket Circuit IR has no generic `if`.
  

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/qasm_context.py
+++ b/qiskit_braket_provider/providers/qasm_context.py
@@ -115,7 +115,6 @@ class _QiskitProgramContext(AbstractProgramContext):
         self._circuit_stack: list[QuantumCircuit] = [QuantumCircuit()]
         self._param_map: dict[str, Parameter] = {}
         self._in_verbatim_box = False
-        self._verbatim_circuit: QuantumCircuit | None = None
         self._verbatim_box_name = verbatim_box_name
         self._clbit_offset: dict[str, int] = {}
 
@@ -218,13 +217,7 @@ class _QiskitProgramContext(AbstractProgramContext):
         active.add_bits([Qubit() for _ in range(num_missing_qubits)])
         self.num_qubits = max(self.num_qubits, active.num_qubits)
 
-        if self._in_verbatim_box:
-            # Ensure verbatim circuit also has enough qubits by adding missing qubits
-            num_missing_qubits = max_qubits - self._verbatim_circuit.num_qubits
-            self._verbatim_circuit.add_bits([Qubit() for _ in range(num_missing_qubits)])
-            self._verbatim_circuit.append(CircuitInstruction(gate, target))
-        else:
-            active.append(CircuitInstruction(gate, target))
+        active.append(CircuitInstruction(gate, target))
 
     def handle_parameter_value(self, value: Number | Expr) -> Number | Parameter:
         return _sympy_to_qiskit(value, self._param_map) if isinstance(value, Expr) else value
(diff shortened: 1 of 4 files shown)


qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -218,13 +217,7 @@ def add_gate_instruction(
...
         self.num_qubits = max(self.num_qubits, active.num_qubits)
 
-        if self._in_verbatim_box:
-            # Ensure verbatim circuit also has enough qubits by adding missing qubits
-            num_missing_qubits = max_qubits - self._verbatim_circuit.num_qubits
-            self._verbatim_circuit.add_bits([Qubit() for _ in range(num_missing_qubits)])
-            self._verbatim_circuit.append(CircuitInstruction(gate, target))
-        else:
-            active.append(CircuitInstruction(gate, target))
+        active.append(CircuitInstruction(gate, target))
```
aniksd-braket, line 220: Why have you removed the verbatim case here? This was used to ensure that when the main circuit is extended to have more qubits, the verbatim circuit is too

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -246,44 +239,23 @@ def add_measure(
...
-        - Reset verbatim state
-
-        Args:
-            marker: VerbatimBoxDelimiter indicating START_VERBATIM or END_VERBATIM
-
-        Raises:
-            ValueError: If nested verbatim boxes are encountered or if END_VERBATIM is called without START_VERBATIM
-        """
-
+        """Handle verbatim box start/end markers by scoping body construction on the stack."""
```
aniksd-braket: can you add the `Args` and `Raises` to the doc string

nilsquet: I adjust the docstring.

qiskit_braket_provider/providers/qasm_context.py
nilsquet: The qubit-sync still happens, just through the stack instead of an explicit branch. Inside a verbatim box, `_active_circuit` returns the body (top of `_circuit_stack`), so the `active.add_bits(...)` call above already adds missing qubits to the body — no separate code path needed. At END_VERBATIM, `_extend_bits(parent, body)` propagates any qubits the body grew back to the parent before the BoxOp is appended.

Covered by test_verbatim_box_adds_qubits_to_main_circuit and test_non_contiguous_physical_qubits.

aniksd-braket (approved):

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -249,41 +242,37 @@ def add_verbatim_marker(self, marker: VerbatimBoxDelimiter) -> None:
...
-            box_op = BoxOp(self._verbatim_circuit, label=self._verbatim_box_name)
-
-            active = self._active_circuit
-            # Append BoxOp to active circuit with all qubits (convert indices to Qubit objects)
-            qubit_objects = [active.qubits[i] for i in range(self._verbatim_circuit.num_qubits)]
-            active.append(box_op, qubit_objects)
-
+            body = self._circuit_stack.pop()
+            parent = self._active_circuit
+            self._extend_bits(parent, body)
```
yitchen-tim, line 273: Is this a right behavior? Shouldn't local variable's liftcycle ends with the existing of the box?
https://openqasm.com/language/scope.html

yitchen-tim: @nilsquet I realized the PR is close during the review. Is the above a real concern? I guess it's not impacting the current use case, so we can parking lot it.

PR #350, opened 2026-06-23
fix: classical bit in box

nilsquet:
*Description of changes:*  
  `_extract_verbatim_boxes` forwarded `clbit_indices` to its `Barrier` placeholder. Qiskit's `Barrier` is qubit-only, so this raised `CircuitError` whenever a verbatim `BoxOp` carried clbits (e.g. a body containing a measurement). The clbit wiring lives in the box body and is restored by `_restore_verbatim_boxes`, so the placeholder doesn't need clbits.
  
  *Testing done:*
  
  New unit test that round-trips a verbatim `BoxOp` with a measurement through `to_braket` and asserts the emitted OpenQASM preserves the clbit declaration and measurement wiring. Fails on mainline with the expected `CircuitError`. `tox -e py313,linters_check` passes.

Example:
```
  from qiskit import QuantumCircuit
  from qiskit_braket_provider.providers.adapter import to_braket
  
  qc = QuantumCircuit(2, 2)
  with qc.box(label="verbatim"):
      qc.h(0)
      qc.cx(0, 1)
      qc.measure([0, 1], [0, 1])
  
  to_braket(qc)
```

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/amazon-braket/qiskit-braket-provider/blob/main/CONTRIBUTING.md) doc
- [x] I used the PR title format described in [CONTRIBUTING](https://github.com/amazon-braket/qiskit-braket-provider/blob/main/CONTRIBUTING.md#pr-title-format)
- [ ] I have updated any necessary documentation, including [READMEs](https://github.com/amazon-braket/qiskit-braket-provider/blob/main/README.md) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have checked that my tests are not configured for a specific region or account (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #356, opened 2026-06-29
infra: replace verbatim box functions with TransformationPasses

aniksd-braket:
## Description:
  
  Replaces _extract_verbatim_boxes and _restore_verbatim_boxes with proper Qiskit TransformationPass subclasses (ExtractVerbatimBoxes and
  RestoreVerbatimBoxes) in a new passes/ module.
  
## Changes:
  
  - Added qiskit_braket_provider/providers/passes/ with ExtractVerbatimBoxes and RestoreVerbatimBoxes
  - _compile now integrates the passes into the PassManager pipeline (pre_init / post_optimization) instead of manual pre/post-processing
  - Handles BoxOps with classical bits (aligned with #350)
  - Removed standalone _extract_verbatim_boxes / _restore_verbatim_boxes from compilation.py
  - Updated tests to exercise the pass-based API
  -  Fixes a correctness bug where _restore_verbatim_boxes on main discarded outer qubit indices, causing verbatim box contents to be placed on [0, 1, ..., k-1] regardless of where the user appended the BoxOp; the new pass-based approach preserves the original qubit mapping, with regression tests for reversed and non-contiguous qubit orderings.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -41,8 +41,6 @@ from qiskit_braket_provider.providers.compilation import (
     _CompilationContext,  # noqa: F401
     _compile,
     _default_target,  # noqa: F401
-    _extract_verbatim_boxes,  # noqa: F401
-    _restore_verbatim_boxes,  # noqa: F401
 )
 from qiskit_braket_provider.providers.gate_mappings import (
     _BRAKET_GATE_NAME_TO_QISKIT_GATE,
@@ -145,7 +143,7 @@ def to_braket(
     basis_gates: Collection[str] | None = None,
     coupling_map: list[list[int]] | None = None,
     angle_restrictions: Mapping[str, Mapping[int, set[float] | tuple[float, float]]] | None = None,
-    optimization_level: int = 0,
+    optimization_level: int | None = 0,
     callback: Callable | None = None,
     num_processes: int | None = None,
     pass_manager: PassManager | None = None,
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -143,7 +143,7 @@ def to_braket(
     basis_gates: Collection[str] | None = None,
     coupling_map: list[list[int]] | None = None,
     angle_restrictions: Mapping[str, Mapping[int, set[float] | tuple[float, float]]] | None = None,
-    optimization_level: int | None = 0,
+    optimization_level: int = 0,
     callback: Callable | None = None,
     num_processes: int | None = None,
     pass_manager: PassManager | None = None,
@@ -182,7 +182,7 @@ def to_braket(
         angle_restrictions (Mapping[str, Mapping[int, set[float] | tuple[float, float]]] | None):
             Mapping of gate names to parameter angle constraints used to
             validate numeric parameters. Default: ``None``.
-        optimization_level (int | None): The optimization level to pass to ``qiskit.transpile``.
+        optimization_level (int): The optimization level to pass to ``qiskit.transpile``.
             From Qiskit:
 
             * 0: no optimization - basic translation, no optimization, trivial layout
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 
 from qiskit import QuantumCircuit, generate_preset_pass_manager
 from qiskit.circuit import Barrier, BoxOp, Measure
-from qiskit.transpiler import PassManager, Target
+from qiskit.transpiler import PassManager, StagedPassManager, Target
 
 from braket.aws import AwsDevice
 from braket.devices import Device
@@ -216,8 +216,12 @@ def _compile(
                 seed_transpiler=seed_transpiler,
             )
             if has_verbatim_boxes:
-                pm.pre_init = PassManager([ExtractVerbatimBoxes(verbatim_box_name)])
-                pm.post_optimization = PassManager([RestoreVerbatimBoxes(verbatim_box_name)])
+                pm = StagedPassManager(
+                    stages=("verbatim_extract", "transpile", "verbatim_restore"),
+                    verbatim_extract=PassManager([ExtractVerbatimBoxes(verbatim_box_name)]),
+                    transpile=pm,
+                    verbatim_restore=PassManager([RestoreVerbatimBoxes(verbatim_box_name)]),
+                )
             circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
         elif has_verbatim_boxes:
             circuits = [_inline_verbatim_boxes(circ, verbatim_box_name) for circ in circuits]
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -64,6 +64,20 @@ def _default_target(circuits: Iterable[QuantumCircuit]) -> Target:
     return target
 
 
+def _inline_verbatim_boxes(circuit: QuantumCircuit, verbatim_box_name: str) -> QuantumCircuit:
+    """Replace BoxOps with the given label by their contents at the top level."""
+    out = circuit.copy_empty_like()
+    for instr in circuit.data:
+        op = instr.operation
+        if isinstance(op, BoxOp) and getattr(op, "label", None) == verbatim_box_name:
+            qubit_idx = [circuit.find_bit(q).index for q in instr.qubits]
+            clbit_idx = [circuit.find_bit(c).index for c in instr.clbits]
+            out.compose(op.blocks[0], qubits=qubit_idx, clbits=clbit_idx, inplace=True)
+        else:
+            out.append(instr.operation, instr.qubits, instr.clbits)
+    return out
+
+
 def _compile(
     circuits: QuantumCircuit | Iterable[QuantumCircuit],
     *,
@@ -193,10 +207,7 @@ def _compile(
             )
         ):
             pm = generate_preset_pass_manager(
-                # generate_preset_pass_manager does not accept None unlike transpile().
-                # If user explicitly passes None, default to 2 to match Qiskit's transpile() behavior.
-                # When not passed, the signature default of 0 is used (no optimization).
-                optimization_level=optimization_level if optimization_level is not None else 2,
+                optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
                 coupling_map=coupling_map,
                 target=target,
@@ -209,19 +220,9 @@ def _compile(
                 pm.post_optimization = PassManager([RestoreVerbatimBoxes(verbatim_box_name)])
             circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
         elif has_verbatim_boxes:
-            # No transpilation needed but still need to extract/restore verbatim boxes
-            verbatim_pm = PassManager([
-                ExtractVerbatimBoxes(verbatim_box_name),
-                RestoreVerbatimBoxes(verbatim_box_name),
-            ])
-            circuits = verbatim_pm.run(circuits)
+            circuits = [_inline_verbatim_boxes(circ, verbatim_box_name) for circ in circuits]
     elif has_verbatim_boxes:
-        # verbatim=True: unpack BoxOps without transpilation
-        verbatim_pm = PassManager([
-            ExtractVerbatimBoxes(verbatim_box_name),
-            RestoreVerbatimBoxes(verbatim_box_name),
-        ])
-        circuits = verbatim_pm.run(circuits)
+        circuits = [_inline_verbatim_boxes(circ, verbatim_box_name) for circ in circuits]
 
     if isinstance(target, _SubstitutedTarget):
         circuits = target._substitute(circuits)
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -7,7 +7,6 @@ and the run() endpoints, including verbatim box handling and transpilation.
 import warnings
 from collections.abc import Callable, Collection, Iterable, Mapping, Sequence
 from dataclasses import dataclass
-from typing import TypeVar
 
 from qiskit import QuantumCircuit, generate_preset_pass_manager
 from qiskit.circuit import Barrier, BoxOp, Measure
@@ -30,12 +29,20 @@ from qiskit_braket_provider.providers.target import (
     local_simulator_to_target,
 )
 
-_T = TypeVar("_T")
-
 
 @dataclass(frozen=True)
 class _CompilationContext:
-    """Internal result from _compile containing compiled circuits and resolved state."""
+    """Result of :func:`_compile` containing compiled circuits and resolved state.
+
+    Attributes:
+        circuits: The compiled Qiskit QuantumCircuits.
+        target: The resolved transpiler target, if any.
+        qubit_labels: Physical qubit indices on the target device.
+        verbatim: Whether verbatim mode was requested.
+        basis_gates: The basis gate set used for compilation.
+        angle_restrictions: Per-gate angle constraints from the device.
+        pass_manager: The custom PassManager used, if any.
+    """
 
     circuits: list[QuantumCircuit]
     target: Target | None
@@ -58,8 +65,8 @@ def _default_target(circuits: Iterable[QuantumCircuit]) -> Target:
 
 
 def _compile(
-    circuits: QuantumCircuit | Iterable[QuantumCircuit] = None,
-    *args,
+    circuits: QuantumCircuit | Iterable[QuantumCircuit],
+    *,
     qubit_labels: Sequence[int] | None = None,
     target: Target | None = None,
     verbatim: bool | None = None,
@@ -77,35 +84,63 @@ def _compile(
     routing_method: str | None = None,
     seed_transpiler: int | None = None,
 ) -> _CompilationContext:
+    """Compile Qiskit circuits for execution on a Braket device.
+
+    This is the compilation pipeline used by :func:`to_braket`. It handles
+    verbatim box extraction/restoration, transpilation via Qiskit's pass infrastructure,
+    and device-specific gate substitutions.
+
+    Args:
+        circuits: One or more Qiskit QuantumCircuits to compile.
+        qubit_labels: Physical qubit indices on the target device. If not supplied,
+            contiguous indices are assumed.
+        target: A Qiskit transpiler target describing device constraints.
+        verbatim: If ``True``, skip transpilation (pass circuits through unchanged).
+        basis_gates: Set of gate names supported by the target device.
+        coupling_map: Qubit connectivity as a list of ``[control, target]`` pairs.
+        angle_restrictions: Per-gate angle constraints from the device.
+        optimization_level: Transpiler optimization level (0-3). Default: 0.
+        callback: Callback function passed to the transpiler.
+        num_processes: Number of parallel processes for transpilation.
+        pass_manager: A custom Qiskit PassManager. Mutually exclusive with other
+            transpilation options.
+        braket_device: A Braket Device to derive target and qubit labels from.
+        connectivity: Deprecated alias for ``coupling_map``.
+        verbatim_box_name: Label identifying verbatim BoxOp nodes.
+        layout_method: Layout method for the transpiler.
+        routing_method: Routing method for the transpiler.
+        seed_transpiler: Seed for reproducible transpilation.
+
+    Returns:
+        A :class:`_CompilationContext` containing the compiled circuits and resolved
+        compilation state.
+
+    Raises:
+        ValueError: If mutually exclusive options are specified together.
+        TypeError: If inputs are not QuantumCircuits.
+    """
     if isinstance(circuits, QuantumCircuit):
         circuits = [circuits]
     circuits = list(circuits)
 
-    if len(args) > 4:
-        raise ValueError(f"Unknown arguments passed: {args[4:]}")
-    padded = args + (None,) * max(0, 4 - len(args))
-    basis_gates = _check_positional(padded[0], basis_gates, "basis_gates")
-    verbatim = _check_positional(padded[1], verbatim, "verbatim")
-    connectivity = _check_positional(padded[2], connectivity, "connectivity")
-    angle_restrictions = _check_positional(padded[3], angle_restrictions, "angle_restrictions")
     _validate_arguments(
         circuits, target, basis_gates, coupling_map, connectivity, pass_manager, braket_device
     )
     coupling_map = coupling_map or connectivity
 
-    has_verbatim_boxes = any(
-        isinstance(instr.operation, BoxOp)
-        and getattr(instr.operation, "label", None) == verbatim_box_name
-        for circ in circuits
-        for instr in circ.data
-    )
+    has_barriers_named_verbatim = False
+    has_verbatim_boxes = False
 
-    if any(
-        isinstance(instr.operation, Barrier)
-        and getattr(instr.operation, "label", None) == verbatim_box_name
-        for circ in circuits
-        for instr in circ.data
-    ):
+    for circ in circuits:
+        for instr in circ.data:
+            label = getattr(instr.operation, "label", None)
+            if label == verbatim_box_name:
+                if isinstance(instr.operation, Barrier):
+                    has_barriers_named_verbatim = True
+                elif isinstance(instr.operation, BoxOp):
+                    has_verbatim_boxes = True
+
+    if has_barriers_named_verbatim:
         raise ValueError(
             "Cannot have a Barrier labeled with the same label used for verbatim boxes"
         )
@@ -200,19 +235,6 @@ def _compile(
         angle_restrictions=angle_restrictions,
         pass_manager=pass_manager,
     )
-
-
-def _check_positional(pos: _T, kw: _T, name: str) -> _T:
-    if pos is None:
-        return kw
-    if kw is not None:
-        raise TypeError(f"Multiple values for {name}: {pos, kw}")
-    warnings.warn(
-        f"Passing {name} as a positional argument is deprecated.",
-        DeprecationWarning,
-        stacklevel=1,
-    )
-    return pos
 
 
 def _validate_arguments(
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -7,8 +7,9 @@ and the run() endpoints, including verbatim box handling and transpilation.
 import warnings
 from collections.abc import Callable, Collection, Iterable, Mapping, Sequence
 from dataclasses import dataclass
+from typing import TypeVar
 
-from qiskit import QuantumCircuit, transpile
+from qiskit import QuantumCircuit, generate_preset_pass_manager
 from qiskit.circuit import Barrier, BoxOp, Measure
 from qiskit.transpiler import PassManager, Target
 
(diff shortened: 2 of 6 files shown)


qiskit_braket_provider/providers/compilation.py
```diff
@@ -294,29 +192,40 @@ def _compile(
...
             )
         ):
-            circuits = transpile(
-                circuits,
-                basis_gates=basis_gates,
+            pm = generate_preset_pass_manager(
+                # generate_preset_pass_manager does not accept None unlike transpile().
+                # If user explicitly passes None, default to 2 to match Qiskit's transpile() behavior.
+                # When not passed, the signature default of 0 is used (no optimization).
+                optimization_level=optimization_level if optimization_level is not None else 2,
```
yitchen-tim: nit: an assert `optimization_level==2` may be better than changing upstream's input

qiskit_braket_provider/providers/compilation.py
yitchen-tim, line 199: Not related for this PR, but just wanted to say this condition is so complicated to understand what actaully goes into if vs else branch

qiskit_braket_provider/providers/compilation.py
```diff
@@ -294,29 +192,40 @@ def _compile(
...
                 layout_method=effective_layout_method,
                 routing_method=effective_routing_method,
                 seed_transpiler=seed_transpiler,
             )
+            if has_verbatim_boxes:
+                pm.pre_init = PassManager([ExtractVerbatimBoxes(verbatim_box_name)])
+                pm.post_optimization = PassManager([RestoreVerbatimBoxes(verbatim_box_name)])
+            circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
+        elif has_verbatim_boxes:
+            # No transpilation needed but still need to extract/restore verbatim boxes
```
yitchen-tim: What difference does it make compared to no pass at all?

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -190,36 +208,41 @@ def test_single_verbatim_box_restoration(h_cx_circuit: QuantumCircuit):
...
+    qc.append(Barrier(NUM_QUBITS, label=VERBATIM_LABEL), QUBIT_PAIR)
+
+    pm = PassManager([ExtractVerbatimBoxes(VERBATIM_LABEL)])
+    with pytest.raises(ValueError, match="conflicts with the verbatim box label"):
+        pm.run(qc)
 
-    with pytest.raises(ValueError, match=error_match):
-        _restore_verbatim_boxes(transpiled, boxes, VERBATIM_LABEL)
+
+def test_restore_raises_on_missing_box():
```
yitchen-tim: nit: it's testing name conflict exception than missing box

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+    pm = PassManager([RestoreVerbatimBoxes(VERBATIM_LABEL)])
+    result = pm.run(qc)
+    assert result.data[0].operation.name == "h"
+
+
+def test_restore_raises_on_mismatched_barrier_label():
+    """RestoreVerbatimBoxes raises when a verbatim barrier has no matching box (line 127-129)."""
+    from qiskit.converters import circuit_to_dag
+
+    from qiskit_braket_provider.providers.passes.verbatim_passes import _indexed_label
```
yitchen-tim: imports should be on top. Something good to add to your Kiro steering doc

qiskit_braket_provider/providers/passes/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Qiskit transpiler passes for the Braket provider."""
+
+from qiskit_braket_provider.providers.passes.verbatim_passes import (
+    ExtractVerbatimBoxes,
+    RestoreVerbatimBoxes,
+)
+
+__all__ = ["ExtractVerbatimBoxes", "RestoreVerbatimBoxes"]
```
nilsquet:
```suggestion
from .verbatim_passes import ExtractVerbatimBoxes as ExtractVerbatimBoxes
from .verbatim_passes import RestoreVerbatimBoxes as RestoreVerbatimBoxes
```

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -0,0 +1,158 @@
...
+  transpiler will leave untouched, and stashes the original circuits in
+  ``property_set["verbatim_boxes"]``.
+
+- :class:`RestoreVerbatimBoxes` (post-transpilation): replaces the labeled
+  barriers with the stashed gate sequences.
+
+Both passes must share a ``property_set``, which happens automatically when
+they live in the same :class:`~qiskit.transpiler.PassManager` or
+:class:`~qiskit.transpiler.StagedPassManager`.
+"""
```
nilsquet: This documentation should be reduced. I don't think this is the right place to educate customers about what verbatim circuits are.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -145,7 +143,7 @@ def to_braket(
     basis_gates: Collection[str] | None = None,
     coupling_map: list[list[int]] | None = None,
     angle_restrictions: Mapping[str, Mapping[int, set[float] | tuple[float, float]]] | None = None,
-    optimization_level: int = 0,
+    optimization_level: int | None = 0,
```
nilsquet: What's the reason for this change? This allows now to explicitly setting `optimization_level=None` which is not a supported parameter type in `_compile`.

qiskit_braket_provider/providers/compilation.py
nilsquet: Why don't we drop the `None` and set the default to `2` [similar to Qiskit](https://github.com/Qiskit/qiskit/blob/9ff63334c989c8c6e0ddd303a224ddbd907f0b6e/qiskit/compiler/transpiler.py#L280)?

qiskit_braket_provider/providers/compilation.py
nilsquet, line 81: This method was already hard to follow before this PR and the added complexity doesn't make it easier. It would be great if we could introduce some helper functions to make it more digestible. This is hard to maintain in the future otherwise.

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -190,36 +208,41 @@ def test_single_verbatim_box_restoration(h_cx_circuit: QuantumCircuit):
...
-    restored = _restore_verbatim_boxes(
-        transpiled, [(h_circuit, QUBIT_PAIR), (cx_circuit, QUBIT_PAIR)], VERBATIM_LABEL
-    )
+    pm = PassManager([ExtractVerbatimBoxes(VERBATIM_LABEL), RestoreVerbatimBoxes(VERBATIM_LABEL)])
+    restored = pm.run(main)
 
     gate_names = [i.operation.name for i in restored.data]
     assert gate_names == ["h", "x", "cx"]
 
+    # Verify qubit indices
```
nilsquet:
```suggestion
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -124,16 +126,21 @@ def test_verbatim_box_extraction(
...
 
-    assert len(boxes) == 1
-    box_circuit, qubit_indices = boxes[0]
+    verbatim_boxes = pm.property_set["verbatim_boxes"]
+    assert len(verbatim_boxes) == 1
+    box_circuit, _clbits = next(iter(verbatim_boxes.values()))
     assert [d.operation.name for d in box_circuit.data] == expected_gate_names
-    assert qubit_indices == expected_qubits
+
+    # Verify the barrier is on the expected qubits
```
nilsquet:
```suggestion
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+
+    body = QuantumCircuit(NUM_QUBITS)
+    body.h(0)
+
+    qc = QuantumCircuit(NUM_QUBITS)
+    qc.append(BoxOp(body, label=VERBATIM_LABEL), QUBIT_PAIR)
+
+    # Provide basis_gates that include all the circuit's gates so no
+    # transpilation is triggered, but verbatim boxes still need handling.
+    # The circuit only has a "box" gate at top level (the BoxOp).
```
nilsquet:
```suggestion
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+        restore_pass.run(dag)
+
+
+def test_restore_raises_on_leftover_boxes():
+    """RestoreVerbatimBoxes raises when boxes remain after processing (line 153)."""
+    from qiskit.converters import circuit_to_dag
+
+    from qiskit_braket_provider.providers.passes.verbatim_passes import _indexed_label
+
+    # Circuit with no verbatim barriers
```
nilsquet:
```suggestion
```

qiskit_braket_provider/providers/compilation.py
nilsquet: The reason is that the box operators must be flattened, both here and in the other `elif has_verbatim_boxes` case. However, this should be done in a helper method and not via executing the verbatim handling passes which have cause this as a side-effect. E.g., via something like 

```python
  def _inline_verbatim_boxes(circuit, label):
      """Replace BoxOps with the given label by their contents at the top level."""
      out = circuit.copy_empty_like()
      for instr in circuit.data:
          op = instr.operation
          if isinstance(op, BoxOp) and getattr(op, "label", None) == label:
              qubit_idx = [circuit.find_bit(q).index for q in instr.qubits]
              clbit_idx = [circuit.find_bit(c).index for c in instr.clbits]
              out.compose(op.blocks[0], qubits=qubit_idx, clbits=clbit_idx, inplace=True)
          else:
              out.append(instr.operation, instr.qubits, instr.clbits)
      return out
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
     assert "cnot $0, $1;" in out
     assert "b[0] = measure $1;" in out
     assert "b[1] = measure $0;" in out
+
+
+def test_verbatim_boxes_without_transpilation_needed():
+    """Cover the elif path: verbatim boxes present but no transpilation triggers."""
+    from qiskit_braket_provider.providers.compilation import _compile
```
nilsquet: Please move to the top.

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+    dag = circuit_to_dag(qc)
+    with pytest.raises(RuntimeError, match="has no matching box"):
+        restore_pass.run(dag)
+
+
+def test_restore_raises_on_leftover_boxes():
+    """RestoreVerbatimBoxes raises when boxes remain after processing (line 153)."""
+    from qiskit.converters import circuit_to_dag
+
+    from qiskit_braket_provider.providers.passes.verbatim_passes import _indexed_label
```
nilsquet: same here

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+    body.h(0)
+    restore_pass.property_set["verbatim_boxes"] = {different_label: (body, [])}
+
+    dag = circuit_to_dag(qc)
+    with pytest.raises(RuntimeError, match="has no matching box"):
+        restore_pass.run(dag)
+
+
+def test_restore_raises_on_leftover_boxes():
+    """RestoreVerbatimBoxes raises when boxes remain after processing (line 153)."""
```
nilsquet: Line numbers must be removed because they are outdated soon.
```suggestion
    """RestoreVerbatimBoxes raises when boxes remain after processing."""
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+
+    # Provide basis_gates that include all the circuit's gates so no
+    # transpilation is triggered, but verbatim boxes still need handling.
+    # The circuit only has a "box" gate at top level (the BoxOp).
+    result = _compile(qc, basis_gates={"h", "cx", "box"})
+    assert result.circuits[0].data[0].operation.name == "h"
+
+
+def test_restore_without_extract_is_noop():
+    """RestoreVerbatimBoxes with empty property_set returns dag unchanged (line 120)."""
```
nilsquet:
```suggestion
    """RestoreVerbatimBoxes with empty property_set returns dag unchanged."""
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+    qc = QuantumCircuit(NUM_QUBITS)
+    qc.h(0)
+
+    pm = PassManager([RestoreVerbatimBoxes(VERBATIM_LABEL)])
+    result = pm.run(qc)
+    assert result.data[0].operation.name == "h"
+
+
+def test_restore_raises_on_mismatched_barrier_label():
+    """RestoreVerbatimBoxes raises when a verbatim barrier has no matching box (line 127-129)."""
```
nilsquet:
```suggestion
    """RestoreVerbatimBoxes raises when a verbatim barrier has no matching box."""
```

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -142,44 +149,55 @@ def test_multiple_verbatim_boxes_extraction(h_circuit: QuantumCircuit, cx_circui
...
 
 
 def test_single_verbatim_box_restoration(h_cx_circuit: QuantumCircuit):
-    transpiled = QuantumCircuit(NUM_QUBITS)
-    transpiled.append(Barrier(NUM_QUBITS, label=VERBATIM_LABEL), QUBIT_PAIR)
+    main = QuantumCircuit(NUM_QUBITS)
+    main.append(BoxOp(h_cx_circuit, label=VERBATIM_LABEL), QUBIT_PAIR)
 
-    restored = _restore_verbatim_boxes(transpiled, [(h_cx_circuit, QUBIT_PAIR)], VERBATIM_LABEL)
+    pm = PassManager([ExtractVerbatimBoxes(VERBATIM_LABEL), RestoreVerbatimBoxes(VERBATIM_LABEL)])
```
nilsquet: I think the purpose of this test has changed. Before, it just tested the restoring.

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -190,36 +208,41 @@ def test_single_verbatim_box_restoration(h_cx_circuit: QuantumCircuit):
...
-    transpiled.append(Barrier(NUM_QUBITS, label=VERBATIM_LABEL), QUBIT_PAIR)
+    main = QuantumCircuit(NUM_QUBITS)
+    main.append(BoxOp(h_circuit, label=VERBATIM_LABEL), QUBIT_PAIR)
+    main.x(1)
+    main.append(BoxOp(cx_circuit, label=VERBATIM_LABEL), QUBIT_PAIR)
 
-    restored = _restore_verbatim_boxes(
-        transpiled, [(h_circuit, QUBIT_PAIR), (cx_circuit, QUBIT_PAIR)], VERBATIM_LABEL
-    )
+    pm = PassManager([ExtractVerbatimBoxes(VERBATIM_LABEL), RestoreVerbatimBoxes(VERBATIM_LABEL)])
```
nilsquet: Same here.

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+
+
+def test_restore_without_extract_is_noop():
+    """RestoreVerbatimBoxes with empty property_set returns dag unchanged (line 120)."""
+    qc = QuantumCircuit(NUM_QUBITS)
+    qc.h(0)
+
+    pm = PassManager([RestoreVerbatimBoxes(VERBATIM_LABEL)])
+    result = pm.run(qc)
+    assert result.data[0].operation.name == "h"
```
nilsquet, line 662: This assert is not sufficient to check what's indicated in the docstring. We should also check that there are no other gates and potentially further checks.

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +563,92 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+
+    restore_pass = RestoreVerbatimBoxes(VERBATIM_LABEL)
+    # Set verbatim_boxes that will never be consumed
+    label = _indexed_label(VERBATIM_LABEL, 0)
+    body = QuantumCircuit(NUM_QUBITS)
+    body.h(0)
+    restore_pass.property_set["verbatim_boxes"] = {label: (body, [])}
+
+    dag = circuit_to_dag(qc)
+    with pytest.raises(RuntimeError, match="verbatim boxes lost during transpilation"):
```
nilsquet: Is this error message really fitting? From the docstring, is sounds like we have too many and not too few boxes remaining.

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -0,0 +1,158 @@
...
+            if not _is_verbatim_label(label, self._verbatim_box_name):
+                continue
+            if label not in verbatim_boxes:
+                raise RuntimeError(
+                    f"Internal error: verbatim barrier '{label}' has no matching box. "
+                    f"This is a bug in the verbatim pass pipeline."
+                )
+            box_circuit, clbit_indices = verbatim_boxes.pop(label)
+            if clbit_indices:
+                # The barrier only has qubits; build a replacement DAG that
```
nilsquet: This comment is hard to understand, since we just checked for classical indices. Maybe rather `build a replacement DAG that includes the box's clbits`?

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -0,0 +1,158 @@
...
+                # includes the box's clbits so we can substitute correctly.
+                box_dag = circuit_to_dag(box_circuit)
+                # Wire mapping: barrier qubits → box qubits, dag clbits → box clbits
+                qubit_map = dict(zip(node.qargs, box_dag.qubits, strict=True))
```
nilsquet: The `wires` dict in `RestoreVerbatimBoxes` is built in the wrong direction. Qiskit's
  [`substitute_node_with_dag`](https://docs.quantum.ibm.com/api/qiskit/qiskit.dagcircuit.DAGCircuit#substitute_node_with_dag) expects `{input_dag wire: outer wire}`, but `dict(zip(node.qargs, box_dag.qubits))` produces `{outer wire: input_dag wire}`.
  
  This crashes with a Rust panic whenever the outer circuit and the body have unequal qubits — e.g. a 2-qubit verbatim `BoxOp` on a subset of a 3-qubit circuit, with measurements inside:
  
  ```python
  body = QuantumCircuit(2, 2)
  body.x(0); body.measure(0, 0); body.measure(1, 1)
  
  qc = QuantumCircuit(3, 2)
  qc.append(BoxOp(body, label="verbatim"), [qc.qubits[0], qc.qubits[1]], qc.clbits)
  
  to_braket(qc, verbatim=True)
  # pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
  
  Existing tests don't catch this because same-sized outer/body anonymous Qubits compare equal across circuits, so the dict direction is
  irrelevant.
  
  Fix — swap the zip argument order:
  
  qubit_map = dict(zip(box_dag.qubits, node.qargs, strict=True))
  clbit_map = dict(zip(
      box_dag.clbits,
      [dag.clbits[i] for i in clbit_indices],
      strict=True,
  ))
  
  Verified end-to-end: crash repros on the PR branch, fix resolves it, full suite stays green (355 passing). Please add a test with mismatched outer/body sizes.

qiskit_braket_provider/providers/compilation.py
aniksd-braket: Changing default to 2 is a breaking change as the current default is 0. This might need some discussion and a separate PR. However, I can definitely drop the `None`

qiskit_braket_provider/providers/compilation.py
aniksd-braket: Will do this in a follow-up PR

aniksd-braket: I agree, will simplify in follow-up PR

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -540,3 +630,82 @@ def test_to_braket_round_trip_preserves_verbatim_box_with_measurement():
...
+    body.h(0)
+    restore_pass.property_set["verbatim_boxes"] = {label: (body, [])}
+
+    dag = circuit_to_dag(qc)
+    with pytest.raises(RuntimeError, match="stashed verbatim boxes were not restored"):
+        restore_pass.run(dag)
+
+
+def test_extract_restore_with_non_verbatim_barrier():
+    """A non-verbatim labeled barrier is preserved through extract/restore."""
```
nilsquet:
```suggestion
    """A non-verbatim barrier is preserved through extract/restore."""
```
It has no label.

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, verbatim_box_name: str = _BRAKET_VERBATIM_BOX_NAME):
+        super().__init__()
+        self._verbatim_box_name = verbatim_box_name
+
+    def run(self, dag: DAGCircuit) -> DAGCircuit:
+        """Splice stashed gate sequences back in place of labeled barriers.
+
+        Raises:
+            RuntimeError: If the number of labeled barriers does not match
```
nilsquet: The second potential `RuntimeError` is not covered here.

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -0,0 +1,136 @@
...
+            if not _is_verbatim_label(label, self._verbatim_box_name):
+                continue
+            if label not in verbatim_boxes:
+                raise RuntimeError(
+                    f"Internal error: verbatim barrier '{label}' has no matching box. "
+                    f"This is a bug in the verbatim pass pipeline."
+                )
+            box_circuit, clbit_indices = verbatim_boxes.pop(label)
+            if clbit_indices:
+                box_dag = circuit_to_dag(box_circuit)
```
nilsquet: Can be hoisted above the`if`, cause the else branch does it as well.

qiskit_braket_provider/providers/compilation.py
```diff
@@ -294,29 +206,27 @@ def _compile(
...
+                coupling_map=coupling_map,
                 target=target,
-                callback=callback,
-                num_processes=num_processes,
                 layout_method=effective_layout_method,
                 routing_method=effective_routing_method,
                 seed_transpiler=seed_transpiler,
             )
+            if has_verbatim_boxes:
+                pm.pre_init = PassManager([ExtractVerbatimBoxes(verbatim_box_name)])
```
nilsquet: Assigning to `pm.pre_init` clobbers Qiskit's `ContainsInstruction` + `ConditionalController` that `generate_preset_pass_manager` sets there. Downstream passes rely on the property-set flags those set. Wrap the preset in a fresh outer `StagedPassManager` instead of overwriting its stages:
  
  ```python
  pm = generate_preset_pass_manager(
      optimization_level=optimization_level,
      basis_gates=list(basis_gates) if basis_gates else None,
      coupling_map=coupling_map,
      target=target,
      layout_method=effective_layout_method,
      routing_method=effective_routing_method,
  )
  if has_verbatim_boxes:
      pm = StagedPassManager(
          stages=("verbatim_extract", "transpile", "verbatim_restore"),
          verbatim_extract=PassManager([ExtractVerbatimBoxes(verbatim_box_name)]),
          transpile=pm,
          verbatim_restore=PassManager([RestoreVerbatimBoxes(verbatim_box_name)]),
      )
  circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
  ```

  Needs StagedPassManager added to the qiskit.transpiler import.
  
  Please add regression tests: (1) transpiling a circuit with a verbatim box through a target should still run Qiskit's ContainsInstruction (verifiable via callback=); (2) a circuit with a verbatim box + an if_test targeting a backend without control-flow support should raise the "control-flow not supported"
  error rather than a BasisTranslator "Unable to translate" error.


Here's an example of the current erroneous behavior:
```python
  from qiskit import transpile
  from qiskit.circuit.library import CXGate, HGate, Measure, XGate
  
  from qiskit_braket_provider.providers.adapter import to_braket, to_qiskit
  
  target = Target(num_qubits=3)
  target.add_instruction(HGate(), {(i,): None for i in range(3)})
  target.add_instruction(XGate(), {(i,): None for i in range(3)})
  target.add_instruction(CXGate(), {(0, 1): None, (1, 2): None})
  target.add_instruction(Measure(), {(i,): None for i in range(3)})
  
  qc = to_qiskit("""
  OPENQASM 3.0;
  bit[1] c;
  #pragma braket verbatim
  box {
      h $0;
      cnot $0, $1;
  }
  c[0] = measure $0;
  if (c[0] == 1) {
      x $2;
  }
  """)
  
  transpile(qc, target=target)   # Qiskit: clean error
  to_braket(qc, target=target)   # QBP:    confusing error
  ```
  Output:
  
  - transpile(qc, target=target) → TranspilerError: "The control-flow constructs ['if_else', 'box'] are not supported by the backend."
  - to_braket(qc, target=target) → TranspilerError: 'Unable to translate the operations in the circuit: {("if_else", 3)} to the backend\'s (or manually specified) target basis: {"h", "x", "cx", "measure", "barrier", "snapshot", "store"}...'

File: test/unit_tests/test_adapter_to_braket_verbatim.py [hunk 27]
```diff
@@ -519,6 +556,59 @@ def test_to_braket_handles_verbatim_box_with_clbits():
...
+    body.x(0)
+    body.measure(0, 0)
+    body.measure(1, 1)
+
+    qc = QuantumCircuit(3, 2)
+    qc.append(BoxOp(body, label=VERBATIM_LABEL), [qc.qubits[0], qc.qubits[1]], qc.clbits)
+
+    bc = to_braket(qc, verbatim=True)
+
+    source = bc.to_ir(ir_type="OPENQASM").source
```
yitchen-tim, line 575: Would be much better to assert the entire OQ3 program. Asserting these lines does not tell us about boxes (e.g., do we expect verbatim box to be in the OQ3 or not etc). Same comment for the two tests below

nilsquet (changes requested):

qiskit_braket_provider/providers/compilation.py
aniksd-braket: Good find, fixed

nilsquet:
This PR silently fixes a real correctness bug on `main` that isn't called out anywhere. `_restore_verbatim_boxes` on `main` discards the outer qubit indices recorded at extract time and uses box-internal indices instead:
  
  ```python
  box_circuit, _ = next(verbatim_box_iter)                                # outer indices discarded
  for box_instruction in box_circuit.data:
      qubit_indices = [box_circuit.find_bit(q).index for q in box_instruction.qubits]  # box-internal
      reconstructed_circuit.append(box_instruction.operation, qubit_indices, ...)      # applied to outer
  ```
  Any verbatim box appended on anything other than [q0, q1, ..., qk] in order gets its contents placed on [0, 1, ..., k-1] regardless of where the user put it. For gates with orientation (cx, ccx, ...) this silently produces a physically different gate.
  
  Repro on main:
  
```python
  body = QuantumCircuit(2); body.cx(0, 1)
  qc = QuantumCircuit(3)
  qc.append(BoxOp(body, label="verbatim"), [qc.qubits[0], qc.qubits[2]])
  out = _compile(qc, basis_gates={"cx"}).circuits[0]
  cx = next(i for i in out.data if i.operation.name == "cx")
  assert [out.find_bit(q).index for q in cx.qubits] == [0, 2]
  # main: got [0, 1]
  # this PR: passes
  ```

  The fix (substitute_node_with_dag(node, box_dag, wires=...)) handles this because node.qargs are the actual outer qubits. No existing test caught it because every verbatim test uses [0, 1] / [0, 1, 2], which coincides with box-internal indices.
  
  Two asks:
  
  1. Add a regression test for non-contiguous / reversed placement:
  ```python
      def test_verbatim_box_preserves_non_contiguous_outer_qubits():
          body = QuantumCircuit(2); body.cx(0, 1)
          qc = QuantumCircuit(3)
          qc.append(BoxOp(body, label=VERBATIM_LABEL), [qc.qubits[0], qc.qubits[2]])
          out = _compile(qc, basis_gates={"cx"}).circuits[0]
          cx = next(i for i in out.data if i.operation.name == "cx")
          assert [out.find_bit(q).index for q in cx.qubits] == [0, 2]
  
      def test_verbatim_box_preserves_reversed_qubit_order():
          body = QuantumCircuit(2); body.cx(0, 1)
          qc = QuantumCircuit(2)
          qc.append(BoxOp(body, label=VERBATIM_LABEL), [qc.qubits[1], qc.qubits[0]])
          out = _compile(qc, basis_gates={"cx"}).circuits[0]
          cx = next(i for i in out.data if i.operation.name == "cx")
          assert [out.find_bit(q).index for q in cx.qubits] == [1, 0]
  ```
  2. Mention the fix in the PR description / release note so users relying on partial verbatim boxes on main know their output will (correctly) change.

nilsquet (approved):

PR #360, opened 2026-07-06
feat: result pragma support

aniksd-braket:
# Add support for Braket result type pragmas
  
  ## Summary
  
  Adds end-to-end support for Braket result type pragmas (`#pragma braket result ...`) in the `to_qiskit()` → `to_braket()` pipeline. Previously, any OpenQASM program containing result pragmas would crash
  with `NotImplementedError`. Now they are parsed, stored as circuit metadata, and processed into the correct basis rotation gates and measurements.
  
  ## Changes
  
  **`qasm_context.py`** — Parse and store result pragmas
  - Override `parse_pragma()` to capture the raw pragma text
  - Override `add_result()` to store parsed IR objects as circuit metadata
  - Attach `braket_result_pragmas` to circuit metadata on access
  
  **`passes/basis_rotation_pass.py`** — New `AddBasisRotationGates` TransformationPass
  - Reads result pragma metadata from the circuit
  - Appends appropriate basis rotation gates per observable:
    - Z, I → no rotation
    - X → H
    - Y → Z, S, H
    - H → Ry(-π/4)
    - Hermitian → Unitary (eigenvector matrix)
  - Appends measurements on all targeted qubits
  - Skips basis-invariant types (state_vector, density_matrix, amplitude)
  
  **`compilation.py`** — Integrate into `_compile()` pipeline
  - Detects circuits with result pragma metadata after transpilation
  - Runs `AddBasisRotationGates` pass automatically
  
  ## Testing
  
  - **16 tests** — Parsing and metadata attachment in `to_qiskit()`
  - **14 tests** — `AddBasisRotationGates` pass unit tests (all observable types, edge cases)
  - **15 tests** — End-to-end through `to_braket()` with full OQ3 string assertions and metadata verification
  
  ## Result type support matrix
  
  | Result Type | Behavior |
  |------------|----------|
  | Expectation, Sample, Variance | Basis rotation gates + measurement |
  | Probability | Measurement only (Z-basis) |
  | StateVector, DensityMatrix, Amplitude | No gates or measurements added |

## Merge Checklist

#### General

#### Tests


qiskit_braket_provider/providers/compilation.py
```diff
@@ -231,6 +237,13 @@ def _compile(
     if isinstance(target, _SubstitutedTarget):
         circuits = target._substitute(circuits)
 
+    has_result_pragmas = any(
+        (circ.metadata or {}).get("braket_result_pragmas") for circ in circuits
+    )
+    if has_result_pragmas and not preserve_result_pragmas:
+        basis_rotation_pm = PassManager([AddBasisRotationGates()])
```
nilsquet, line 244: This compilation pass must run before the actual compilation. Right now, 1) the added rotation gates are not nativized and 2) the applied mapping is not considered.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+            return dag
+
+        num_qubits = dag.num_qubits()
+        qubits_to_measure = set()
+        rotation_ops = []
+
+        for pragma_entry in result_pragmas:
+            parsed = pragma_entry["parsed"]
+
+            if isinstance(parsed, _BASIS_INVARIANT_TYPES):
```
nilsquet, line 97: `state_vector`, `density_matrix`, and `amplitude` don't need rotation or measurement, so skipping them here is right. But the pragma only lives on `circuit.metadata["braket_result_pragmas"]` at this point, and nothing downstream (qasm3.dumps, to_braket) reads that metadata back. The result is OQ3 output with no pragma and no measurement — the simulator has no result type to compute.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+    try:
+        np_matrix = np.array([[complex(c[0], c[1]) for c in row] for row in matrix])
+    except (IndexError, TypeError, ValueError) as e:
+        raise ValueError(f"Invalid Hermitian matrix format in result type pragma: {e}") from e
+    _eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
+    unitary = eigenvectors.conj().T
+    return [(UnitaryGate(unitary), targets)]
+
+
+class AddBasisRotationGates(TransformationPass):
```
nilsquet, line 64: The naming is misleading since this function also adds measurements.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+                            qubits_to_measure.update(obs_targets)
+                            obs_idx += num_qubits_for_obs
+
+        if not qubits_to_measure and not rotation_ops:
+            return dag
+
+        num_clbits_needed = len(qubits_to_measure)
+        clbits_to_add = num_clbits_needed - dag.num_clbits()
+        if clbits_to_add > 0:
+            for _ in range(clbits_to_add):
```
nilsquet, line 141: If the source already had bit[N] c; or explicit measure, the pass reuses those clbits and adds new measurements on top, producing silent double-measurement.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+                dag.add_clbits([Clbit()])
+
+        for gate, target in rotation_ops:
+            if isinstance(target, list):
+                dag.apply_operation_back(gate, [dag.qubits[t] for t in target])
+            else:
+                dag.apply_operation_back(gate, [dag.qubits[target]])
```
nilsquet, line 148: target and qubit here originate from parsed.targets (source-level indices like q[0]→0). After transpile, dag.qubits[i] is the qubit at DAG slot i, not the source's q[i]. Rotations and measurements land on the wrong qubit whenever the layout permutes.

Should be fine if this pass is executed before the compilation.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+    Variance,
+)
+
+_BASIS_INVARIANT_TYPES = (StateVector, DensityMatrix, Amplitude)
+_Z_BASIS_TYPES = (Probability,)
+_OBSERVABLE_TYPES = (Expectation, Sample, Variance)
+
+
+def _rotation_gates_for_observable(observable: str, target: int) -> list[tuple]:
+    """Return (gate, qubit) pairs for rotating from the given observable basis to Z basis."""
```
nilsquet, line 27: Inconsistent docstring style compared to `_rotation_gates_for_hermitian` where args and returns is explicitly stated.

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -132,6 +135,41 @@ def circuit(self) -> QuantumCircuit:
             )
         return self._circuit_stack[0]
 
+    def parse_pragma(self, pragma_body: str):  # noqa: ANN202
```
nilsquet, line 138: Why is `# noqa: ANN202` needed? Adding a return type should fix it. Also, the docstring should document what's returned.

qiskit_braket_provider/providers/qasm_context.py
```diff
@@ -132,6 +135,41 @@ def circuit(self) -> QuantumCircuit:
...
+
+        Overrides AbstractProgramContext.parse_pragma() to store the raw pragma
+        command string before parsing. This allows us to reconstruct the full
+        pragma text for re-emission in the compiled output.
+
+        Args:
+            pragma_body: The body of the pragma statement (e.g., "braket result expectation z(q[0])").
+        """
+        self._last_pragma_command = pragma_body
+        return super().parse_pragma(pragma_body)
```
nilsquet, line 149: `parse_pragma` override and `_last_pragma_command` field only exist to smuggle raw text over to `add_result` so it can build `raw_pragma`. But `raw_pragma` isn't consumed anywhere in this PR — it's written to metadata and never read. Drop both.

If a future consumer needs the pragma text, reconstruct it from `result` at that point (`braket.circuits.result_types.*` classes already implement `_to_openqasm`).

Bonus: the `"adjoint_gradient" in self._last_pragma_command` substring check becomes `isinstance(result, AdjointGradient)` — precise, and no longer needs the raw text.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+    Expectation,
+    Probability,
+    Sample,
+    StateVector,
+    Variance,
+)
+
+_BASIS_INVARIANT_TYPES = (StateVector, DensityMatrix, Amplitude)
+_Z_BASIS_TYPES = (Probability,)
+_OBSERVABLE_TYPES = (Expectation, Sample, Variance)
```
nilsquet, line 23: Should this be moved to `gate_mappings.py` where we store all other constants? Admittedly, the file name is not suitable and should rather be called something like "constants" or "common".

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,156 @@
...
+    measurements are added. For observable types (expectation, sample, variance),
+    rotation gates are added before measurements.
+
+    AdjointGradient is not supported by this pass — it requires the pragma to be
+    preserved verbatim for simulator-side processing. Programs containing
+    adjoint_gradient pragmas are not parseable by the default simulator's pragma
+    grammar and will not reach this pass.
+    """
+
+    def run(self, dag: DAGCircuit) -> DAGCircuit:
```
nilsquet, line 82: The pass does a lot in one method: type dispatch (basis-invariant / Z-basis / observable), target resolution (None → all), observable-form dispatch (single string vs. list vs. hermitian matrix), tensor-product bookkeeping via obs_idx, deferred measurement collection, clbit allocation, and DAG mutation.
  Each branch is small on its own, but the interactions are where the subtle behavior lives.

Suggest splitting: a small pure helper per result-type that returns a (rotation_gates, qubits_to_measure) plan from a parsed object, and a thin pass body that unions plans and applies them to the DAG. Easier to reason about and to test.

nilsquet (changes requested):
Thanks for the thorough work here — two high-level notes on the test suite:

1. Feels heavy: ~900 lines of tests for ~200 lines of source, with each result type exercised at the parse, pass, and end-to-end layers. Suggest collapsing to end-to-end per result type plus a small unit-test file for edge cases. Given the volume, I didn't review the test bodies line-by-line.
2. to_braket is always called without `target`, `braket_device`, or `basis_gates`. Therefore, the test don't show that the added rotation gates are not compiled.

qiskit_braket_provider/providers/qasm_context.py
laurencap: +1 to this. It's always possible to add back later if it ends up being needed elsewhere.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
aniksd-braket: Yup, that's a valid concern. I will make sure to make sure that the metadata is appropriately passed to downstream simulators and devices.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
aniksd-braket: added to guard to make sure a circuit with result pragmas does not have pre-existing measurements

laurencap:
This has been broken into smaller PRs which have been merged! 🎉  Closing.

PR #364, opened 2026-07-13
fix: barrier scope and bare-barrier handling in to_qiskit

nilsquet:
*Description of changes:*

  Fix two bugs in the Braket-`Circuit` input path of `to_qiskit` (`adapter.py`), both in the `case compiler_directives.Barrier():` match arm:
  - **Verbatim routing.** Barriers between `StartVerbatimBox` and `EndVerbatimBox` were written to the outer `qiskit_circuit` instead of the `verbatim_buffer` that becomes the `BoxOp` body. Now route to `verbatim_buffer` when active, matching the routing pattern the loop already uses for every other gate (`verbatim_buffer is None` dispatch).
  - **Global barrier dropped.** After amazon-braket/amazon-braket-sdk-python#1203, `Circuit.barrier()` (bare) emits `Barrier([])` with an empty
  target — a late-binding "all qubits" marker. The old code produced `qiskit_circuit.barrier([])`, a zero-qubit no-op. Now fall back to `active.qubits` when `instruction.target` is empty.
  
  *Testing done:*
  Added three unit tests in `test/unit_tests/test_adapter.py`.
  
  Full suite: `tox` clean; `test_adapter.py` 99 passed, 12 barrier tests passed, `ruff` clean.
## Merge Checklist

#### General

#### Tests

### DIFF
--- a/test/unit_tests/test_adapter.py
+++ b/test/unit_tests/test_adapter.py
@@ -1747,7 +1747,7 @@ class TestFromBraket(TestCase):
 
     def test_braket_circuit_bare_barrier_inside_verbatim_box_covers_body_qubits(self):
         """Bare barrier inside a verbatim box covers the box body's qubits."""
-        circuit = Circuit().h(0).add_verbatim_box(Circuit().h(0).barrier().cnot(0, 1)).h(1)
+        circuit = Circuit().add_verbatim_box(Circuit().h(0).barrier().cnot(0, 1))
         qc = to_qiskit(circuit, add_measurements=False)
         box = next(i for i in qc.data if i.operation.name == "box").operation
         body = box.body
--- a/test/unit_tests/test_adapter.py
+++ b/test/unit_tests/test_adapter.py
@@ -1710,6 +1710,52 @@ class TestFromBraket(TestCase):
         self.assertEqual({0, 1}, barrier_indices[0])
         self.assertEqual({0}, barrier_indices[1])
 
+    def test_braket_circuit_barrier_scoping_around_verbatim_box(self):
+        """Barriers land in the scope they were written in."""
+        circuit = (
+            Circuit()
+            .h(0)
+            .barrier([0])
+            .add_verbatim_box(Circuit().h(0).barrier([0, 1]).cnot(0, 1))
+            .barrier([1])
+            .h(1)
+        )
+        qc = to_qiskit(circuit, add_measurements=False)
+        ops = [(i.operation.name, [qc.find_bit(q).index for q in i.qubits]) for i in qc.data]
+        self.assertEqual(
+            ops,
+            [
+                ("h", [0]),
+                ("barrier", [0]),
+                ("box", [0, 1]),
+                ("barrier", [1]),
+                ("h", [1]),
+            ],
+        )
+        body = qc.data[2].operation.body
+        body_ops = [
+            (i.operation.name, [body.find_bit(q).index for q in i.qubits]) for i in body.data
+        ]
+        self.assertEqual(body_ops, [("h", [0]), ("barrier", [0, 1]), ("cx", [0, 1])])
+
+    def test_braket_circuit_bare_barrier_covers_all_qubits(self):
+        """Bare Circuit.barrier() (empty target) covers every qubit."""
+        circuit = Circuit().h(0).barrier().h(1)
+        qc = to_qiskit(circuit, add_measurements=False)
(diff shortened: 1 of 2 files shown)


test/unit_tests/test_adapter.py
```diff
@@ -1710,6 +1710,63 @@ def test_braket_barrier_to_qiskit(self):
...
+        qc = to_qiskit(circuit, add_measurements=False)
+        box = next(i for i in qc.data if i.operation.name == "box").operation
+        body = box.body
+        body_ops = [
+            (i.operation.name, [body.find_bit(q).index for q in i.qubits]) for i in body.data
+        ]
+        self.assertEqual(body_ops, [("h", [0]), ("barrier", [0, 1]), ("cx", [0, 1])])
+
+    def test_braket_circuit_bare_barrier_in_verbatim_box_with_qubit_outside_box(self):
+        """Bare barrier inside a verbatim box expands to all outer-circuit qubits."""
```
sesmart, line 1760: Is this intended behavior? That it is allocated to the entire circuit?

test/unit_tests/test_adapter.py
```diff
@@ -1710,6 +1710,63 @@ def test_braket_barrier_to_qiskit(self):
...
+    def test_braket_circuit_bare_barrier_covers_all_qubits(self):
+        """Bare Circuit.barrier() (empty target) covers every qubit."""
+        circuit = Circuit().h(0).barrier().h(1)
+        qc = to_qiskit(circuit, add_measurements=False)
+        ops = [(i.operation.name, [qc.find_bit(q).index for q in i.qubits]) for i in qc.data]
+        self.assertEqual(ops, [("h", [0]), ("barrier", [0, 1]), ("h", [1])])
+
+    def test_braket_circuit_bare_barrier_inside_verbatim_box_covers_body_qubits(self):
+        """Bare barrier inside a verbatim box covers the box body's qubits."""
+        circuit = Circuit().h(0).add_verbatim_box(Circuit().h(0).barrier().cnot(0, 1)).h(1)
```
sesmart: Maybe remove the h(0) and h(1) around the verbatim box? Or else this one and the next test the same thing - that the barrier is inferred from the total qubits of the entire circuit.

test/unit_tests/test_adapter.py
nilsquet: Yes, this is how I read the [OQ3 spec](https://openqasm.com/language/delays.html#:~:text=A%20barrier%20can%20also%20be%20invoked%20without%20arguments%2C%20in%20which%20case%20the%20argument%20is%20assumed%20to%20be%20all%20qubits).

test/unit_tests/test_adapter.py
nilsquet: This test verifies that the barrier includes a following higher-indexed qubit inside the verbatim box, while the other verifies it for a higher-indexed qubit outside the box.

sesmart: Right, so for this test if it is testing inside the box, do you need the external h(0) & h(1) ?

sesmart: (Specifically, h(1), which sets the outside circuit as (0,1) )

nilsquet: True. I removed them.

sesmart: Thank you!

sesmart (approved):

test/unit_tests/test_adapter.py
```diff
@@ -1710,6 +1710,63 @@ def test_braket_barrier_to_qiskit(self):
...
+        qc = to_qiskit(circuit, add_measurements=False)
+        box = next(i for i in qc.data if i.operation.name == "box").operation
+        body = box.body
+        body_ops = [
+            (i.operation.name, [body.find_bit(q).index for q in i.qubits]) for i in body.data
+        ]
+        self.assertEqual(body_ops, [("h", [0]), ("barrier", [0, 1]), ("cx", [0, 1])])
+
+    def test_braket_circuit_bare_barrier_in_verbatim_box_with_qubit_outside_box(self):
+        """Bare barrier inside a verbatim box expands to all outer-circuit qubits."""
```
yitchen-tim, line 1760: It does not really test "verbatim box expands to all outer-circuit qubits" because outer circuit in this case has no qubits. (I assume outer circuit means outside of verbatim box.)

yitchen-tim (approved):

PR #362, opened 2026-07-09
feat: changes in the context to parse result pragmas

aniksd-braket:
## Summary
  
  Extends `_QiskitProgramContext` to parse and store Braket result type pragmas (`#pragma braket result ...`) as structured metadata on the Qiskit
  `QuantumCircuit`.
  
  This is Part 1 of a multi-PR effort to support end-to-end result pragma compilation through the `to_qiskit()` → `to_braket()` pipeline. The larger PR still exists as #360 
  
  ## Changes
  
  - Override `parse_pragma()` in `_QiskitProgramContext` to capture the raw pragma string before delegating to the base class parser
  - Override `add_result()` to collect parsed result IR objects and attach them as circuit metadata under
  `circuit.metadata["braket_result_pragmas"]`
  - Each entry contains `{"raw_pragma": str, "parsed": Results}` for downstream consumption
  - Reject `adjoint_gradient` result type with a clear `NotImplementedError` (not supported in this pipeline)
  - Add 17 unit tests covering all supported result types (expectation, probability, sample, variance, state_vector, amplitude, density_matrix),
  metadata structure validation, and error handling
  
  ## Supported result types
  
  | Result Type | Stored in metadata |
  |---|---|
  | `expectation` | ✅ |
  | `probability` | ✅ |
  | `sample` | ✅ |
  | `variance` | ✅ |
  | `state_vector` | ✅ |
  | `amplitude` | ✅ |
  | `density_matrix` | ✅ |
  | `adjoint_gradient` | ❌ (raises NotImplementedError) |
  
  ## Testing
  
  - 17 new tests in `test/unit_tests/test_adapter_to_qiskit_result_pragmas.py`
  - All existing tests continue to pass
  
  ## Next PRs
  
  - Part 2: `AddBasisRotationGates` transpiler pass (converts metadata → rotation gates + measurements)
  - Part 3: Integration into `_compile()` with `preserve_result_pragmas` flag

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [ ] I have read the [CONTRIBUTING](https://github.com/amazon-braket/qiskit-braket-provider/blob/main/CONTRIBUTING.md) doc
[...]

PR #366, opened 2026-07-14
feat: add basis rotation pass

aniksd-braket:
## Summary
  
  Adds the `AddBasisRotationAndMeasurement` transpiler pass that converts result type pragma metadata into basis rotation gates and measurements on the circuit.
  
  This is Part 2 of a multi-PR effort to support end-to-end result pragma compilation through the `to_qiskit()` → `to_braket()` pipeline. Depends on Part 1 (PR #362).
  
  ## Changes
  
  - **`passes/basis_rotation_pass.py`** — New `AddBasisRotationAndMeasurement` TransformationPass
    - Reads `braket_result_pragmas` from circuit metadata
    - Appends appropriate basis rotation gates per observable:
      - Z, I → no rotation
      - X → H
      - Y → Z, S, H
      - H → Ry(-π/4)
      - Hermitian → Unitary(eigenvectors†)
    - Appends measurements on all targeted qubits
    - Skips basis-invariant types (state_vector, density_matrix, amplitude)
    - Validates observable/target count mismatches and hermitian matrix dimensions
    - Raises `ValueError` if circuit already contains measurements
    - Raises `ValueError` for unrecognized result types
  
  - **`gate_mappings.py`** — Result type classification constants
    - `_BASIS_INVARIANT_RESULT_TYPES` (StateVector, DensityMatrix, Amplitude)
    - `_Z_BASIS_RESULT_TYPES` (Probability)
    - `_OBSERVABLE_RESULT_TYPES` (Expectation, Sample, Variance)
  
  - **`passes/__init__.py`** — Exports `AddBasisRotationAndMeasurement`
  
  ## Design
  
  The pass is structured as pure helper functions that compute a "plan" (rotation ops + qubits to measure) per result type, with a thin `run()` method that unions plans and applies them to the DAG:
  
  - `_rotation_gates_for_observable()` — single-qubit rotation lookup
  - `_rotation_gates_for_hermitian()` — eigenvector decomposition for custom matrices
  - `_plan_for_z_basis_type()` — returns qubits to measure (no rotation needed)
  - `_plan_for_observable_type()` — handles broadcast, tensor product, and hermitian cases
  
  This pass should run **before** transpilation so that added rotation gates are compiled to the device's native gate set.
  
  ## Testing
  
  27 tests covering:
  - All observable types (z, i, x, y, h) with exact gate + qubit + clbit assertions
  - Expectation, Sample, Variance dispatch
  - Tensor products, broadcast, targets=None
  - Probability (Z-basis, measurement only)
  - Basis-invariant types (no-op)
  - Hermitian observable (Unitary gate)
  - Error cases: double measurement, non-power-of-2 matrix, observable/target mismatch, unrecognized type
  
  ## Next PR
  
  - Part 3: Integration into `_compile()` with `preserve_result_pragmas` flag and result type re-emission for simulator targets

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/passes/__init__.py
+++ b/qiskit_braket_provider/providers/passes/__init__.py
@@ -1,4 +1,5 @@
 """Qiskit transpiler passes for the Braket provider."""
 
+from .basis_rotation_pass import AddBasisRotationAndMeasurement as AddBasisRotationAndMeasurement
 from .verbatim_passes import ExtractVerbatimBoxes as ExtractVerbatimBoxes
 from .verbatim_passes import RestoreVerbatimBoxes as RestoreVerbatimBoxes
--- a/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
+++ b/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
@@ -122,9 +122,7 @@ def _plan_for_observable_type(
         for obs in observable:
             if isinstance(obs, str):
                 if obs_idx >= len(targets):
-                    raise ValueError(
-                        "More observables than target qubits in result type pragma."
-                    )
+                    raise ValueError("More observables than target qubits in result type pragma.")
                 target = targets[obs_idx]
                 rotation_ops.extend(_rotation_gates_for_observable(obs, target))
                 qubits_to_measure.add(target)
@@ -134,18 +132,14 @@ def _plan_for_observable_type(
                     raise ValueError(f"Hermitian matrix size {len(obs)} is not a power of 2.")
                 num_qubits_for_obs = int(math.log2(len(obs)))
                 if obs_idx + num_qubits_for_obs > len(targets):
-                    raise ValueError(
-                        "More observables than target qubits in result type pragma."
-                    )
+                    raise ValueError("More observables than target qubits in result type pragma.")
                 obs_targets = targets[obs_idx : obs_idx + num_qubits_for_obs]
                 rotation_ops.extend(_rotation_gates_for_hermitian(obs, obs_targets))
                 qubits_to_measure.update(obs_targets)
                 obs_idx += num_qubits_for_obs
 
         if obs_idx != len(targets):
-            raise ValueError(
-                "Fewer observables than target qubits in result type pragma."
-            )
+            raise ValueError("Fewer observables than target qubits in result type pragma.")
 
     return rotation_ops, qubits_to_measure
 
--- a/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
+++ b/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
@@ -1,6 +1,7 @@
 """Transpiler pass to add basis rotation gates and measurements for Braket result type pragmas."""
 
 import math
+from collections.abc import Hashable
 
 import numpy as np
 from qiskit.circuit import Clbit, Gate, Measure
@@ -16,6 +17,7 @@ from braket.ir.jaqcd import (
 )
 from qiskit_braket_provider.providers.gate_mappings import (
     _BASIS_INVARIANT_RESULT_TYPES,
+    _IDENTITY_BASIS,
     _OBSERVABLE_RESULT_TYPES,
     _OBSERVABLE_TO_BASIS,
     _Z_BASIS_RESULT_TYPES,
@@ -44,6 +46,18 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
         matrix.reshape([2] * n_q * 2),
         list(range(n_q))[::-1] + list(range(n_q, 2 * n_q))[::-1],
     ).reshape((2**n_q, 2**n_q))
+
+
+def _hermitian_key(matrix: list) -> Hashable:
+    """Create a hashable key from a Hermitian matrix for deduplication.
+
+    Args:
+        matrix: Nested list representing the Hermitian matrix in [real, imag] format.
+
+    Returns:
+        A hashable tuple of complex elements for identity comparison.
+    """
+    return tuple(complex(c[0], c[1]) for row in matrix for c in row)
 
 
 def _rotation_gates_for_observable(observable: str, target: int) -> list[RotationOp]:
@@ -115,35 +129,54 @@ def _qubits_for_z_basis_type(result: Probability, num_qubits: int) -> set[int]:
     return set(range(num_qubits))
 
 
-def _check_basis_conflicts(
-    qubit_bases: dict[int, str], new_qubits: set[int], new_basis: str
-) -> None:
-    """Check and record basis assignments, raising on conflicts.
+def _check_basis_conflict(
+    qubit_bases: dict[int, Hashable], qubit: int, basis_key: Hashable
+) -> bool:
+    """Check and record a single qubit's basis assignment, raising on conflicts.
 
-    Z-basis (from Probability or z/i observables) is compatible with other Z-basis
-    assignments but conflicts with any other basis on the same qubit.
+    Same basis on the same qubit is deduplicated (returns False). Different basis
+    on the same qubit raises ValueError. New qubit assignment returns True.
+    Identity ("i") is compatible with any basis and does not override it.
 
     Args:
-        qubit_bases: Mutable dict mapping qubit index to its committed basis.
-        new_qubits: Set of qubit indices being assigned.
-        new_basis: The basis being assigned to these qubits.
+        qubit_bases: Mutable dict mapping qubit index to its committed basis key.
+        qubit: The qubit index being assigned.
+        basis_key: Hashable key identifying the observable basis. "i" means identity
+            (compatible with anything). For Pauli observables this is the string from
+            _OBSERVABLE_TO_BASIS. For Hermitian observables this is a tuple of the
+            matrix elements.
+
+    Returns:
+        True if the qubit was newly committed, False if it already had the same basis
+        or the new basis is identity.
 
     Raises:
-        ValueError: If a qubit already has a different non-compatible basis assigned.
+        ValueError: If the qubit already has a different non-identity basis assigned.
     """
-    for qubit in new_qubits:
-        existing = qubit_bases.get(qubit)
-        if existing is not None and existing != new_basis:
-            raise ValueError(
-                f"Conflicting measurement bases on qubit {qubit}: "
-                f"'{existing}' and '{new_basis}' cannot be measured simultaneously."
-            )
-        qubit_bases[qubit] = new_basis
+    if qubit not in qubit_bases:
+        qubit_bases[qubit] = basis_key
+        return True
+
+    existing = qubit_bases[qubit]
+
+    # Identity is diagonal in every basis (never conflicts); same basis deduplicates
+    if basis_key == _IDENTITY_BASIS or existing == basis_key:
+        return False
+
+    # Upgrade from identity to a real basis
+    if existing == _IDENTITY_BASIS:
+        qubit_bases[qubit] = basis_key
+        return True
+
+    raise ValueError(
+        f"Conflicting measurement bases on qubit {qubit}: "
+        f"cannot measure simultaneously with different observables."
+    )
 
 
 def _plan_for_observable_type(
     result: Expectation | Sample | Variance, num_qubits: int
-) -> tuple[list[RotationOp], dict[int, str]]:
+) -> tuple[list[RotationOp], dict[int, Hashable]]:
     """Compute the rotation/measurement plan for an observable result type.
 
     Args:
@@ -152,7 +185,7 @@ def _plan_for_observable_type(
 
     Returns:
         Tuple of (rotation_ops, qubit_bases) where qubit_bases maps qubit index
-        to its measurement basis string.
+        to a hashable basis key identifying the observable.
 
     Raises:
         ValueError: If a hermitian matrix size is not a power of 2, or if
@@ -162,7 +195,7 @@ def _plan_for_observable_type(
     targets = result.targets if result.targets is not None else list(range(num_qubits))
 
     rotation_ops: list[RotationOp] = []
-    qubit_bases: dict[int, str] = {}
+    qubit_bases: dict[int, Hashable] = {}
 
     if len(observable) == 1 and isinstance(observable[0], str):
         basis = _OBSERVABLE_TO_BASIS.get(observable[0].lower(), observable[0].lower())
@@ -187,8 +220,9 @@ def _plan_for_observable_type(
                     raise ValueError("More observables than target qubits in result type pragma.")
                 obs_targets = targets[obs_idx : obs_idx + num_qubits_for_obs]
                 rotation_ops.extend(_rotation_gates_for_hermitian(obs, obs_targets))
+                h_key = _hermitian_key(obs)
                 for t in obs_targets:
-                    qubit_bases[t] = "hermitian"
+                    qubit_bases[t] = h_key
                 obs_idx += num_qubits_for_obs
 
         if obs_idx != len(targets):
@@ -215,10 +249,14 @@ class AddBasisRotationAndMeasurement(TransformationPass):
     This pass should run before transpilation so that added rotation gates are
     compiled to the device's native gate set and qubit layout is applied correctly.
 
+    If multiple result pragmas target the same qubit with the same observable, rotation
+    gates are applied only once (deduplicated). If different observables target the same
+    qubit, the pass raises ValueError since they cannot be measured simultaneously.
+
     Raises:
         ValueError: If the circuit already contains measurement operations.
         ValueError: If two result pragmas require different measurement bases on
-            the same qubit.
+            the same qubit (non-commuting observables).
         NotImplementedError: If basis-invariant result types (state_vector,
             density_matrix, amplitude) are present, as these are not yet supported
             end-to-end.
@@ -253,7 +291,7 @@ class AddBasisRotationAndMeasurement(TransformationPass):
 
         num_qubits = dag.num_qubits()
         all_rotation_ops: list[RotationOp] = []
-        all_qubit_bases: dict[int, str] = {}
+        all_qubit_bases: dict[int, Hashable] = {}
 
         for result in result_pragmas:
             if isinstance(result, _BASIS_INVARIANT_RESULT_TYPES):
@@ -264,14 +302,27 @@ class AddBasisRotationAndMeasurement(TransformationPass):
 
             if isinstance(result, _Z_BASIS_RESULT_TYPES):
                 qubits = _qubits_for_z_basis_type(result, num_qubits)
-                _check_basis_conflicts(all_qubit_bases, qubits, "z")
+                for qubit in qubits:
+                    _check_basis_conflict(all_qubit_bases, qubit, "z")
                 continue
 
             if isinstance(result, _OBSERVABLE_RESULT_TYPES):
                 rotation_ops, qubit_bases = _plan_for_observable_type(result, num_qubits)
+                new_qubits: set[int] = set()
                 for qubit, basis in qubit_bases.items():
-                    _check_basis_conflicts(all_qubit_bases, {qubit}, basis)
-                all_rotation_ops.extend(rotation_ops)
+                    if _check_basis_conflict(all_qubit_bases, qubit, basis):
+                        new_qubits.add(qubit)
+                # Only add rotation ops targeting qubits not yet rotated
+                for gate, target in rotation_ops:
+                    gate_qubits = set(target) if isinstance(target, list) else {target}
+                    if gate_qubits <= new_qubits:
+                        all_rotation_ops.append((gate, target))
+                    elif gate_qubits & new_qubits:
+                        raise ValueError(
+                            "Multi-qubit observable partially overlaps with previously "
+                            "committed qubits. Cannot apply rotation to a subset of "
+                            "an entangled observable's targets."
+                        )
                 continue
 
             raise ValueError(f"Unrecognized result type: {type(result).__name__}")
--- a/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
+++ b/qiskit_braket_provider/providers/passes/basis_rotation_pass.py
@@ -21,31 +21,10 @@ from qiskit_braket_provider.providers.gate_mappings import (
     _OBSERVABLE_RESULT_TYPES,
     _OBSERVABLE_TO_BASIS,
     _Z_BASIS_RESULT_TYPES,
+    _reverse_endianness,
 )
 
 RotationOp = tuple[Gate, int | list[int]]
-
-
-def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
-    """Reverse qubit endianness of a matrix (Braket big-endian to Qiskit little-endian).
-
-    For single-qubit matrices this is a no-op. For multi-qubit matrices, the tensor
-    factor ordering is reversed so q[0] becomes LSB (Qiskit convention) instead of
-    MSB (Braket convention).
-
-    Args:
-        matrix: Square matrix of dimension 2^n x 2^n.
-
-    Returns:
-        Matrix with reversed qubit ordering.
-    """
-    n_q = int(np.log2(matrix.shape[0]))
-    if n_q <= 1:
-        return matrix
-    return np.transpose(
-        matrix.reshape([2] * n_q * 2),
-        list(range(n_q))[::-1] + list(range(n_q, 2 * n_q))[::-1],
-    ).reshape((2**n_q, 2**n_q))
 
 
 def _hermitian_key(matrix: list) -> Hashable:
@@ -86,7 +65,9 @@ def _rotation_gates_for_observable(observable: str, target: int) -> list[Rotatio
             raise ValueError(f"Unknown observable '{observable}' in result type pragma.")
 
 
-def _rotation_gates_for_hermitian(matrix: list, targets: list[int]) -> list[RotationOp]:
+def _rotation_gates_for_hermitian(
+    matrix: list, targets: list[int]
+) -> tuple[list[RotationOp], np.ndarray]:
     """Compute rotation gates for a Hermitian observable from its eigenvectors.
 
     The resulting unitary is endianness-corrected from Braket's big-endian convention
@@ -98,7 +79,8 @@ def _rotation_gates_for_hermitian(matrix: list, targets: list[int]) -> list[Rota
         targets: Qubit indices the unitary should be applied to.
 
     Returns:
-        List of (gate, targets) tuples.
+        Tuple of (rotation_ops, eigenvalues) where eigenvalues are in ascending order
+        as returned by np.linalg.eigh. Measurement bit i corresponds to eigenvalue[i].
 
     Raises:
         ValueError: If the matrix cannot be parsed or is not a valid Hermitian matrix.
@@ -108,9 +90,9 @@ def _rotation_gates_for_hermitian(matrix: list, targets: list[int]) -> list[Rota
     except (IndexError, TypeError, ValueError) as e:
         raise ValueError(f"Invalid Hermitian matrix format in result type pragma: {e}") from e
     np_matrix = _reverse_endianness(np_matrix)
-    _eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
+    eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
     unitary = eigenvectors.conj().T
-    return [(UnitaryGate(unitary), targets)]
+    return [(UnitaryGate(unitary), targets)], eigenvalues
 
 
 def _qubits_for_z_basis_type(result: Probability, num_qubits: int) -> set[int]:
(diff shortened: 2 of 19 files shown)


qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+            return [(HGate(), target)]
+        case "y":
+            return [(ZGate(), target), (SGate(), target), (HGate(), target)]
+        case "h":
+            return [(RYGate(-math.pi / 4), target)]
+        case _:
+            raise ValueError(f"Unknown observable '{observable}' in result type pragma.")
+
+
+def _rotation_gates_for_hermitian(matrix: list, targets: list[int]) -> list[RotationOp]:
```
yitchen-tim: Does Braket actually support measuring in arbitrary Hermitian basis with expectation value? Just checking

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+    Reads ``braket_result_pragmas`` from the circuit metadata and appends
+    the appropriate rotation gates to change from the observable's eigenbasis
+    to the computational (Z) basis, followed by measurements on all targeted qubits.
+
+    For basis-invariant result types (state_vector, density_matrix, amplitude),
+    no gates or measurements are added. For Z-basis types (probability), only
+    measurements are added. For observable types (expectation, sample, variance),
+    rotation gates are added before measurements.
+
+    This pass should run before transpilation so that added rotation gates are
```
yitchen-tim, line 239: Question: What if users request verbatim mode? I assume basis ration will not work. Is that intended behavior?

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+                continue
+
+            raise ValueError(f"Unrecognized result type: {type(result).__name__}")
+
+        if not all_qubits_to_measure and not all_rotation_ops:
+            return dag
+
+        num_clbits_needed = len(all_qubits_to_measure)
+        clbits_to_add = num_clbits_needed - dag.num_clbits()
+        if clbits_to_add > 0:
```
yitchen-tim: Shouldn't this be an exception?

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+        List of (gate, targets) tuples.
+
+    Raises:
+        ValueError: If the matrix cannot be parsed or is not a valid Hermitian matrix.
+    """
+    try:
+        np_matrix = np.array([[complex(c[0], c[1]) for c in row] for row in matrix])
+    except (IndexError, TypeError, ValueError) as e:
+        raise ValueError(f"Invalid Hermitian matrix format in result type pragma: {e}") from e
+    _eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
```
nilsquet: np_matrix here is in Braket's big-endian layout (target[0] = MSB), but its eigenvectors
  end up in UnitaryGate, which Qiskit treats as little-endian (q0 = LSB). Mismatch swaps
  tensor factors — for ≥2 qubits the rotation diagonalizes the wrong observable.
  
  Repro with Z ⊗ X (Z on target[0], X on target[1]), in the [real, imag] format this
  function accepts:
  
      zx = [
          [[0, 0], [1, 0], [0, 0], [ 0, 0]],
          [[1, 0], [0, 0], [0, 0], [ 0, 0]],
          [[0, 0], [0, 0], [0, 0], [-1, 0]],
          [[0, 0], [0, 0], [-1, 0], [0, 0]],
      ]
      [(gate, _)] = _rotation_gates_for_hermitian(zx, [0, 1])
      # gate.to_matrix() diagonalizes Z⊗X in Braket's basis, but Qiskit places it on
      # [q0, q1] with q0 = LSB, so it ends up diagonalizing X⊗Z instead.
  
  Fix on this line, before eigh:
  
      np_matrix = _reverse_endianness(np_matrix)
  
  Every other Braket→Qiskit UnitaryGate construction in the repo already does this (see
  `_create_qiskit_unitary` in adapter.py). Existing Hermitian tests are all 2×2
  (endianness-invariant); add a ≥2-qubit case not symmetric under qubit swap.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+            raise ValueError(
+                "Circuit already contains measurements. Cannot add result type "
+                "measurements on top of existing ones."
+            )
+
+        num_qubits = dag.num_qubits()
+        all_rotation_ops: list[RotationOp] = []
+        all_qubits_to_measure: set[int] = set()
+
+        for result in result_pragmas:
```
nilsquet, line 287: This loop unions rotation gates and measurement targets across all pragmas without
  checking whether two pragmas request different bases on the same qubit. Non-commuting
  observables can't be measured together — the pass should reject that, not silently
  apply both rotations.
  
  Repro:
  
      qc = QuantumCircuit(2)
      qc.metadata = {"braket_result_pragmas": [
          Expectation(observable=["x"], targets=[0]),
          Expectation(observable=["y"], targets=[0]),
      ]}
      PassManager([AddBasisRotationAndMeasurement()]).run(qc)
      # On q[0]: H  (X→Z), then Z, S, H  (Y→Z), then one measurement.
      # Neither X nor Y is measured.
  
  Same failure with Probability(all) + Expectation("x", [0]): the X rotation corrupts
  the probability distribution on q[0].
  
  The pass is exported as public API, so callers can apply it to circuits they built
  themselves — there's no upstream validator in that path. Track a per-qubit committed
  basis while planning and raise ValueError on conflict (Z/I is compatible with
  Probability).

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+        for gate, target in all_rotation_ops:
+            if isinstance(target, list):
+                dag.apply_operation_back(gate, [dag.qubits[t] for t in target])
+            else:
+                dag.apply_operation_back(gate, [dag.qubits[target]])
+
+        # Measurements are added in sorted qubit order, mapping qubit N to
+        # classical bit index based on position in the sorted set. E.g., if
+        # targets are {3, 1}, measurements will be: q[1]→c[0], q[3]→c[1].
+        for idx, qubit in enumerate(sorted(all_qubits_to_measure)):
```
nilsquet: The behavior is documented in the inline comment above, but not as a public contract.
  Sorted-qubit-index is a defensible choice, but it's not what Braket's native Probability
  or Sample would produce — those index bits by target-list position. Whatever consumes
  these measurements has to reverse-engineer this ordering.
  
  - Move the contract into the class docstring so it's part of the public API.
  - Add a test with non-monotonic targets (e.g. targets=[3, 1]) so a future change to
    the ordering is a loud failure — existing tests all use ascending targets, which
    makes sorted-qubit-index and target-list order indistinguishable.
  - Consider writing the qubit → clbit map to dag.metadata so downstream re-emission
    can consume it directly instead of recomputing it from a sort.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+                all_qubits_to_measure.update(qubits)
+                continue
+
+            raise ValueError(f"Unrecognized result type: {type(result).__name__}")
+
+        if not all_qubits_to_measure and not all_rotation_ops:
+            return dag
+
+        num_clbits_needed = len(all_qubits_to_measure)
+        clbits_to_add = num_clbits_needed - dag.num_clbits()
```
nilsquet: The pre-existing-Measure guard above handles the "explicit measure" case, but not "bit[N] c; declared without measurements" — those clbits are still silently reused.
  
  Case A — source declared more clbits than needed:
  
      qc = QuantumCircuit(2, 3)   # bit[3] c;
      qc.metadata = {"braket_result_pragmas": [Probability(targets=[0, 1])]}
      PassManager([AddBasisRotationAndMeasurement()]).run(qc)
      # measurements land on the user's c[0], c[1].
  
  Case B — source declared fewer than needed:
  
      qc = QuantumCircuit(2, 1)   # bit[1] c;
      # first measurement collides with user's c[0], second lands on a fresh clbit.
  
  Fix: always allocate fresh clbits for pragma measurements and measure into those:
  
      pragma_clbits = [Clbit() for _ in range(len(all_qubits_to_measure))]
      dag.add_clbits(pragma_clbits)
      for idx, qubit in enumerate(sorted(all_qubits_to_measure)):
          dag.apply_operation_back(Measure(), [dag.qubits[qubit]], [pragma_clbits[idx]])
  
  Existing clbits stay untouched; pragma outputs live in the last N clbits.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+                "Circuit already contains measurements. Cannot add result type "
+                "measurements on top of existing ones."
+            )
+
+        num_qubits = dag.num_qubits()
+        all_rotation_ops: list[RotationOp] = []
+        all_qubits_to_measure: set[int] = set()
+
+        for result in result_pragmas:
+            if isinstance(result, _BASIS_INVARIANT_RESULT_TYPES):
```
nilsquet, line 288: Basis-invariant result types (state_vector, density_matrix, amplitude) are still on
  the silent-data-loss path — restating my #360 comment. This pass correctly leaves
  them alone, but nothing after it (qasm3.dumps, to_braket) reads
  `braket_result_pragmas`, so a user who requests state_vector ends up with an OQ3
  program that asks for nothing and a simulator that returns nothing.
  
  Just noting that this PR doesn't move the needle on that. Fine to defer to Part 3,
  but please either land Part 3 before this becomes reachable end-to-end, or have this
  pass raise NotImplementedError for basis-invariant types in the meantime so the
  failure is loud rather than silent.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+        ValueError: If a hermitian matrix size is not a power of 2, or if
+            there are more observables than target qubits.
+    """
+    observable = result.observable
+    targets = result.targets if result.targets is not None else list(range(num_qubits))
+
+    rotation_ops: list[RotationOp] = []
+    # All targeted qubits are measured, including identity observables.
+    # This matches Braket SDK behavior where measurement results are returned
+    # for every qubit in the target list regardless of the observable type.
```
nilsquet: Which line does the comment explain?

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    """Helper to create a circuit with result pragma metadata."""
+    qc = QuantumCircuit(num_qubits)
+    qc.h(0)
+    if num_qubits > 1:
+        qc.cx(0, 1)
+    qc.metadata = {"braket_result_pragmas": pragmas}
+    return qc
+
+
+def _run_pass(qc: QuantumCircuit) -> QuantumCircuit:
```
nilsquet: Nit: 
```suggestion
def _run_pragma_handling_pass(qc: QuantumCircuit) -> QuantumCircuit:
```

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    """Pass raises ValueError for invalid configurations."""
+    if match == "already contains measurements":
+        qc = QuantumCircuit(num_qubits, 1)
+        qc.h(0)
+        qc.measure(0, 0)
+        qc.metadata = {"braket_result_pragmas": pragmas}
+    else:
+        qc = _circuit_with_result_pragmas(num_qubits, pragmas)
+
+    pm = PassManager([AddBasisRotationAndMeasurement()])
```
nilsquet: You should use the helper function here as well.

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    """Multi-qubit, broadcast, and multi-result-type cases produce correct operations."""
+    qc = _circuit_with_result_pragmas(num_qubits, pragmas)
+
+    result = _run_pass(qc)
+    added = _get_ops_after_base(result, 2)
+
+    assert added == expected_ops
+
+
+def test_hermitian_observable_applies_unitary():
```
nilsquet, line 250: Only asserts added[0][0] == "unitary" and added[0][1] == [0]. The unitary matrix itself is never checked. A returned garbage matrix still passes as long as it's a UnitaryGate on qubit 0. Add: build the observable in Qiskit's little-endian basis, sandwich U @ observable @ U.conj().T, assert diagonal. Also add a ≥2-qubit case (Z⊗X or similar) — currently only 2×2 Y is tested.

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    ids=[
+        "tensor_product_x_y",
+        "probability_explicit_targets",
+        "probability_all_qubits",
+        "observable_all_qubits",
+        "single_observable_broadcast",
+        "multiple_result_types",
+    ],
+)
+def test_multi_qubit_cases(pragmas: list, num_qubits: int, expected_ops: list):
```
nilsquet, line 239: Two cases worth adding here:
  
   - Non-monotonic targets, e.g. Probability(targets=[3, 1]) on ≥4 qubits. Every case
     in this block uses ascending targets, so the qubit → clbit mapping (currently
     sorted-qubit order) is untested against target-list order. A regression flipping
     the two would slide.
   - One multi-qubit hermitian in a tensor product, e.g.
     Expectation(observable=["x", 4x4_hermitian], targets=[0, 1, 2]). Covers both the
     mixed pauli + hermitian branch and the multi-target hermitian accounting
     (`obs_idx += num_qubits_for_obs` with step > 1).

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    [
+        [StateVector()],
+        [DensityMatrix(targets=[0, 1])],
+        [Amplitude(states=["00", "11"])],
+        [],
+        None,
+    ],
+    ids=["state_vector", "density_matrix", "amplitude", "empty_pragmas", "no_metadata"],
+)
+def test_no_ops_added(pragmas: list):
```
nilsquet: This test asserts what the pass didn't add, never what it preserved. Verified: the
  pass currently keeps `braket_result_pragmas` in metadata for basis-invariant types,
  but nothing here defends that. A refactor that dropped it would still pass.
  
  Add:
      assert result.metadata["braket_result_pragmas"] == pragmas
  
  Also, no test anywhere covers a mixed case like [StateVector(), Probability(targets=[0])].

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+@pytest.mark.parametrize(
+    "pragmas",
+    [
+        [StateVector()],
+        [DensityMatrix(targets=[0, 1])],
+        [Amplitude(states=["00", "11"])],
+        [],
+        None,
+    ],
+    ids=["state_vector", "density_matrix", "amplitude", "empty_pragmas", "no_metadata"],
```
nilsquet: `no_metadata` id is misleading — fixture always sets the key, so pragmas=None becomes
  {"braket_result_pragmas": None}, not "no metadata."
  
  Also redundant with `empty_pragmas`: both `[]` and `None` hit the same early return.
  
  Drop the None case, or skip setting the key in the fixture when pragmas is None to
  actually test the metadata-absent path.

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,306 @@
...
+    qc.metadata = {"braket_result_pragmas": pragmas}
+    return qc
+
+
+def _run_pass(qc: QuantumCircuit) -> QuantumCircuit:
+    """Run the AddBasisRotationAndMeasurement pass on a circuit."""
+    return PassManager([AddBasisRotationAndMeasurement()]).run(qc)
+
+
+def _get_ops_after_base(
```
nilsquet, line 39: The `2` here matches the fixture's `h + cx` count. Any fixture change silently
  misaligns every call site — pass-added ops get counted as base.
  
  Simpler than plumbing base_count around: capture it in each test right before the
  pass runs.
  
      qc = _circuit_with_result_pragmas(...)
      base_count = len(qc.data)
      result = _run_pass(qc)
      added = _get_ops_after_base(result, base_count)

qiskit_braket_provider/providers/passes/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Qiskit transpiler passes for the Braket provider."""
 
+from .basis_rotation_pass import AddBasisRotationAndMeasurement as AddBasisRotationAndMeasurement
```
laurencap, line 3: I see now this isn't the only line doing this, but `as` allows you to alias an imported value. Why `as` to the same exact name?

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,233 @@
...
+    try:
+        np_matrix = np.array([[complex(c[0], c[1]) for c in row] for row in matrix])
+    except (IndexError, TypeError, ValueError) as e:
+        raise ValueError(f"Invalid Hermitian matrix format in result type pragma: {e}") from e
+    _eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
+    unitary = eigenvectors.conj().T
+    return [(UnitaryGate(unitary), targets)]
+
+
+def _plan_for_z_basis_type(result: Probability, num_qubits: int) -> set[int]:
```
laurencap: This returns a set of qubits. Wouldn't `qubits` for z basis type be a better name? Why "plan"?

laurencap (dismissed):

qiskit_braket_provider/providers/passes/__init__.py
aniksd-braket: ruff F401 requires it for re-exports

test/unit_tests/test_basis_rotation_pass.py
aniksd-braket: NA, I am now raising a NotImplemented for these types. Will address in part 3

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
aniksd-braket: Basis rotation should ideally work in verbatim mode too. This will be handled in part 3 as a subsequent PR to this one

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
aniksd-braket: Yes, it appears so - https://docs.aws.amazon.com/braket/latest/developerguide/braket-openqasm-computing-gradients.html, https://docs.aws.amazon.com/braket/latest/developerguide/braket-openqasm-supported-features.html#braket-openqasm-supported-features-datatypes

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,296 @@
...
+                qubits = _qubits_for_z_basis_type(result, num_qubits)
+                _check_basis_conflicts(all_qubit_bases, qubits, "z")
+                continue
+
+            if isinstance(result, _OBSERVABLE_RESULT_TYPES):
+                rotation_ops, qubit_bases = _plan_for_observable_type(result, num_qubits)
+                for qubit, basis in qubit_bases.items():
+                    _check_basis_conflicts(all_qubit_bases, {qubit}, basis)
+                all_rotation_ops.extend(rotation_ops)
+                continue
```
nilsquet, line 322: Same-label pragmas on the same qubit stack rotations. _check_basis_conflicts accepts same-label as compatible, then all_rotation_ops.extend(rotation_ops) appends unconditionally — so a second X-observable on q0 adds a second H, and H·H = I leaves the basis rotated wrong. Every Hermitian is labeled "hermitian" (line 191), so two different Hermitians on the same qubit also pass and stack.
  
  Repro:
  ```
  qc = QuantumCircuit(1); qc.h(0)
  qc.metadata = {"braket_result_pragmas": [
      Sample(observable=["x"], targets=[0]),
      Expectation(observable=["x"], targets=[0]),
  ]}
  PassManager([AddBasisRotationAndMeasurement()]).run(qc)
  # emits: h, h, h, measure  →  ⟨X⟩ ≈ 0 instead of 1
  ```
  Decide the contract for repeated Hermitians (dedupe by matrix identity vs. always raise) and pin it with tests.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,347 @@
...
+    Returns:
+        Tuple of (rotation_ops, qubit_bases) where qubit_bases maps qubit index
+        to a hashable basis key identifying the observable.
+
+    Raises:
+        ValueError: If a hermitian matrix size is not a power of 2, or if
+            there are more observables than target qubits.
+    """
+    observable = result.observable
+    targets = result.targets if result.targets is not None else list(range(num_qubits))
```
nilsquet, line 179: Duplicate targets in the broadcast branch silently stack rotations. Expectation(observable=["y"], targets=[0, 0]) emits Z, S, H, Z, S, H on qubit 0 — (HSZ)² is not a Y→Z rotation. X looks especially benign — the pass emits H, H, H·H = I, so the qubit is measured in the Z basis while the user asked for X.
  Verified on the current branch: on a state with ⟨Y⟩ = +1, the emitted rotations give P(bit 0) = 0.5 → readout ⟨Y⟩ ≈ 0. Same shape of bug for X, Y, H, and any custom Hermitian; only Z and I hide it (empty rotation stack).
  
  Suggested fix, right after this line — reject rather than silently dedup, and place before the broadcast/tensor-product/Hermitian branches so it covers all three uniformly:
  ```
  if len(set(targets)) != len(targets):
      raise ValueError(
          f"All targets in a result-type pragma must be unique: {targets}"
      )
  ```
  Please pin with a test using Y (or any observable with a nonzero rotation stack) that asserts the raise.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Raises:
+        ValueError: If the matrix cannot be parsed or is not a valid Hermitian matrix.
+    """
+    try:
+        np_matrix = np.array([[complex(c[0], c[1]) for c in row] for row in matrix])
+    except (IndexError, TypeError, ValueError) as e:
+        raise ValueError(f"Invalid Hermitian matrix format in result type pragma: {e}") from e
+    np_matrix = _reverse_endianness(np_matrix)
+    _eigenvalues, eigenvectors = np.linalg.eigh(np_matrix)
```
nilsquet: The Pauli and Hermitian rotation paths use different bit-to-eigenvalue conventions, and this line discards the information a consumer would need to interpret Hermitian measurements.
  
  - Pauli path: hardcoded gate sequence produces U · P · U† = Z, so bit 0 → +1, bit 1 → −1.
  - Hermitian path: np.linalg.eigh returns eigenvalues in ascending order, and the _ here throws them away. Same operator, Y written as a Hermitian, gives U · Y · U† = diag(−1, +1) — bit 0 → −1.
  ```
  import numpy as np
  Y = np.array([[0, -1j], [1j, 0]])
  
  # Pauli 'y' path (hardcoded Z, S, H)
  H = np.array([[1,1],[1,-1]]) / np.sqrt(2); S = np.diag([1, 1j]); Z = np.diag([1,-1])
  U_pauli = H @ S @ Z
  # U Y U† = diag(+1, -1)
  
  # Hermitian 'y' path
  w, v = np.linalg.eigh(Y);  U_herm = v.conj().T
  # w = [-1, +1] (ascending),  U Y U† = diag(-1, +1)
  ```
  Nothing reads braket_pragma_qubit_to_clbit yet, so no runtime bug today — but Part 3 needs to translate bit outcomes into eigenvalues, and it can't do that for Hermitian observables without knowing the eigenvalues that came out of this eigh. The two paths don't have to end up with the same bit-index →
  eigenvalue map (a per-pragma tracked mapping works just as well as a shared convention), but this line is the only place the Hermitian eigenvalues exist, and once it returns they're gone. Please decide whether to canonicalize or track, and pin the contract with a test that measures the same operator via "y" and via the Y Hermitian matrix and asserts the intended downstream interpretation.

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,347 @@
...
+    _IDENTITY_BASIS,
+    _OBSERVABLE_RESULT_TYPES,
+    _OBSERVABLE_TO_BASIS,
+    _Z_BASIS_RESULT_TYPES,
+)
+
+RotationOp = tuple[Gate, int | list[int]]
+
+
+def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
```
nilsquet: _reverse_endianness in providers/adapter.py:643 is byte-for-byte the same math as this one (same reshape, same axis permutation). Import it instead of duplicating:
  ```
 from qiskit_braket_provider.providers.adapter import _reverse_endianness

test/unit_tests/test_basis_rotation_pass.py
```diff
@@ -0,0 +1,514 @@
...
+        "multiple_result_types",
+        "non_monotonic_targets",
+        "compatible_z_basis_no_conflict",
+        "identity_compatible_with_x",
+        "duplicate_observable_deduplicates",
+        "partial_overlap_deduplicates_per_qubit",
+        "empty_pragmas_no_ops",
+    ],
+)
+def test_multi_qubit_cases(pragmas: list, num_qubits: int, expected_ops: list):
```
nilsquet, line 239: The _hermitian_key machinery was added specifically to distinguish "same Hermitian on same qubit → dedup" from "different Hermitians → raise". The tests cover different-Hermitians-conflict (different_hermitians_same_qubit) and same-Pauli-dedup (duplicate_observable_deduplicates), but not
  same-Hermitian-dedup. Please add a case so a future edit to _hermitian_key or the dedup branch can't silently regress:
  ```
  (
      [Sample(observable=[x_mat], targets=[0]),
       Expectation(observable=[x_mat], targets=[0])],
      2,
      [("unitary", [0], []), ("measure", [0], [0])],
  ),

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
aniksd-braket: Thank you for this callout. I will be taking no action this PR, but will add it to part 3

aniksd-braket: tracking metadata in this PR, consumption will be in part 3

qiskit_braket_provider/providers/passes/basis_rotation_pass.py
```diff
@@ -0,0 +1,344 @@
...
+                for qubit in qubits:
+                    _check_basis_conflict(all_qubit_bases, qubit, "z")
+                continue
+
+            if isinstance(result, _OBSERVABLE_RESULT_TYPES):
+                rotation_ops, qubit_bases, eigenvalues = _plan_for_observable_type(
+                    result, num_qubits
+                )
+                if eigenvalues is not None:
+                    pragma_eigenvalues[pragma_idx] = eigenvalues
```
nilsquet: eigenvalues is a scalar reassigned each iteration, so _plan_for_observable_type returns only the last Hermitian factor's array. rotation_ops accumulates via .extend, so the circuit is correct; only metadata is lossy.
  ```
  h1 = [[[0, 0], [1, 0]], [[1, 0], [0, 0]]]   # X, eigenvalues [-1, +1]
  h2 = [[[2, 0], [0, 0]], [[0, 0], [5, 0]]]   # diag(2, 5), eigenvalues [2, 5]
  qc = QuantumCircuit(2)
  qc.metadata = {"braket_result_pragmas": [
      Expectation(observable=[h1, h2], targets=[0, 1]),
  ]}
  PassManager([AddBasisRotationAndMeasurement()]).run(qc).metadata["braket_pragma_eigenvalues"]
  # {0: array([2., 5.])}   ← H1's [-1, +1] dropped
  ```
  Key the eigenvalue metadata by qubit instead of pragma index. _check_basis_conflict already enforces one basis per qubit, so {qubit: eigenvalues} is unambiguous, and Part 3 can translate the virtual indices through TranspileLayout at consumption time the same way Qiskit interprets shots against transpiled circuits. Please pin the fix with a regression test using two distinct Hermitian factors on different qubits.

nilsquet (approved):
LGTM, thanks for your effort on this PR. 

Please add an issue or a task to rename "gate_mappings.py" as a follow-up PR.

PR #375, opened 2026-07-22
fix: resolve panic due to missing clbits on verbatim placeholder

aniksd-braket:
## Summary
  
  Fixes a `PanicException` (`Option::unwrap() on None`) in `RestoreVerbatimBoxes` when compiling circuits containing verbatim `BoxOp` nodes with
  classical bits through either the `target=` or `basis_gates=` transpile path.
  
  ## Root Cause
  
  `ExtractVerbatimBoxes` replaced `BoxOp` nodes (which carry both qubits and clbits) with a `Barrier` placeholder. However, Qiskit's `Barrier`
  cannot carry classical bits. When `RestoreVerbatimBoxes` later called `substitute_node_with_dag(node, box_dag, wires={qubit_map + clbit_map})`,
  the Rust implementation panicked because it tried to route clbit wires through a node that had no clbit edges.
  
  ## Fix
  
  Introduce `VerbatimPlaceholder` — a custom `Instruction` subclass with `_directive = True` that can carry both qubits and clbits. It uses the name
  `'barrier'` so Qiskit's transpiler auto-allows it on both `target=` and `basis_gates=` paths without explicit registration. On restoration,
  `substitute_node_with_dag` succeeds because the placeholder node has matching qubit and clbit edges.
  
  ## Trigger Conditions (all required)
  
  1. A verbatim `BoxOp` with classical bits attached (arises from OQ3 `bit[N] c;` at top-level scope + `#pragma braket verbatim box{...}`)
  2. Compilation via `_compile()` with either `target=` or `basis_gates=` (any optimization level)
  
  ## Testing
  
  - Added parametrized regression test covering both `target=` and `basis_gates=` paths
  - Added mid-circuit measurement ordering test (both paths)
  - All 46 verbatim tests pass; full test suite has zero new failures
  
  ## Reproducer
  
  ```python
  from qiskit_braket_provider import to_qiskit, to_braket
  
  source = """OPENQASM 3.0;
  bit[2] c;
  qubit[2] q;
  h q[0];
  #pragma braket verbatim
  box { rx(0.5) q[0]; cz q[0], q[1]; }
  c[0] = measure q[0];
  c[1] = measure q[1];
  """
  
  qc = to_qiskit(source)
  bc = to_braket(qc, basis_gates=["rx", "cz", "h", "measure"])  # was: PanicException
```

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -210,7 +210,9 @@ def _compile(
             )
         ):
             if has_verbatim_boxes and target and "barrier" not in target.operation_names:
-                target.add_instruction(VerbatimPlaceholder(1, 0), name="barrier")
+                target.add_instruction(
+                    VerbatimPlaceholder(1, 0, label="verbatim"), name="barrier"
+                )
             pm = generate_preset_pass_manager(
                 optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -23,9 +23,6 @@ from qiskit_braket_provider.providers.passes import (
     ExtractVerbatimBoxes,
     RestoreVerbatimBoxes,
 )
-from qiskit_braket_provider.providers.passes.verbatim_passes import (
-    VerbatimPlaceholder,
-)
 from qiskit_braket_provider.providers.target import (
     _SubstitutedTarget,
     aws_device_to_target,
@@ -209,8 +206,6 @@ def _compile(
                 )
             )
         ):
-            if has_verbatim_boxes and target and "barrier" not in target.operation_names:
-                target.add_instruction(VerbatimPlaceholder(1, 0, label="verbatim"), name="barrier")
             pm = generate_preset_pass_manager(
                 optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -210,9 +210,7 @@ def _compile(
             )
         ):
             if has_verbatim_boxes and target and "barrier" not in target.operation_names:
-                target.add_instruction(
-                    VerbatimPlaceholder(1, 0, label="verbatim"), name="barrier"
-                )
+                target.add_instruction(VerbatimPlaceholder(1, 0, label="verbatim"), name="barrier")
             pm = generate_preset_pass_manager(
                 optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -209,13 +209,8 @@ def _compile(
                 )
             )
         ):
-            if has_verbatim_boxes and target:
-                # Register the VerbatimPlaceholder with the target so the
-                # transpiler treats it as a known directive. On the basis_gates
-                # path this isn't needed since VerbatimPlaceholder uses name
-                # 'barrier' which is auto-allowed.
-                if "barrier" not in target.operation_names:
-                    target.add_instruction(VerbatimPlaceholder(1, 0), name="barrier")
+            if has_verbatim_boxes and target and "barrier" not in target.operation_names:
+                target.add_instruction(VerbatimPlaceholder(1, 0), name="barrier")
             pm = generate_preset_pass_manager(
                 optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
--- a/qiskit_braket_provider/providers/compilation.py
+++ b/qiskit_braket_provider/providers/compilation.py
@@ -23,6 +23,9 @@ from qiskit_braket_provider.providers.passes import (
     ExtractVerbatimBoxes,
     RestoreVerbatimBoxes,
 )
+from qiskit_braket_provider.providers.passes.verbatim_passes import (
+    VerbatimPlaceholder,
+)
 from qiskit_braket_provider.providers.target import (
     _SubstitutedTarget,
     aws_device_to_target,
@@ -206,6 +209,13 @@ def _compile(
                 )
             )
         ):
+            if has_verbatim_boxes and target:
+                # Register the VerbatimPlaceholder with the target so the
+                # transpiler treats it as a known directive. On the basis_gates
+                # path this isn't needed since VerbatimPlaceholder uses name
+                # 'barrier' which is auto-allowed.
+                if "barrier" not in target.operation_names:
+                    target.add_instruction(VerbatimPlaceholder(1, 0), name="barrier")
             pm = generate_preset_pass_manager(
                 optimization_level=optimization_level,
                 basis_gates=list(basis_gates) if basis_gates else None,
--- a/qiskit_braket_provider/providers/passes/verbatim_passes.py
+++ b/qiskit_braket_provider/providers/passes/verbatim_passes.py
@@ -13,7 +13,7 @@ def _indexed_label(base: str, index: int) -> str:
 
 
 def _is_verbatim_label(label: str | None, base: str) -> bool:
-    """Check if a label is a verbatim barrier label (base or indexed)."""
+    """Check if a label is a verbatim placeholder label (base or indexed)."""
     if label is None:
         return False
     return label == base or (label.startswith(f"{base}__") and label[len(base) + 2 :].isdigit())
@@ -30,8 +30,7 @@ class VerbatimPlaceholder(Instruction):
     _directive = True
 
     def __init__(self, num_qubits: int, num_clbits: int, label: str):
-        super().__init__("barrier", num_qubits, num_clbits, [])
-        self.label = label
+        super().__init__("barrier", num_qubits, num_clbits, [], label=label)
 
 
 class ExtractVerbatimBoxes(TransformationPass):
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: qiskit_braket_provider/providers/passes/verbatim_passes.py [hunk 1]
```diff
@@ -19,6 +19,23 @@ def _is_verbatim_label(label: str | None, base: str) -> bool:
     return label == base or (label.startswith(f"{base}__") and label[len(base) + 2 :].isdigit())
 
 
+class VerbatimPlaceholder(Instruction):
+    """A directive instruction used as a placeholder for verbatim boxes with clbits.
+
+    Uses the name ``'barrier'`` so the transpiler's basis translator automatically
+    allows it without needing explicit registration in the target or basis_gates.
+    Distinguished from real barriers by its label (set to the indexed verbatim label).
```
laurencap: Distinguished from barriers by label, but label is optional? Maybe label should be required to avoid ambiguity.

According to `barrier = Barrier(len(node.qargs), label=label)`, barriers also take labels. I feel like this comment is misleading.

Remember: less is more. No comment is always better than a misleading comment. How does this matter to the user? Does it actually matter?

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -59,14 +76,13 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
-                # so replace with a sub-DAG containing only the barrier on qubits.
-                qc = QuantumCircuit(len(node.qargs), len(node.cargs))
-                qc.append(barrier, list(range(len(node.qargs))))
-                dag.substitute_node_with_dag(node, circuit_to_dag(qc))
+                # Use a VerbatimPlaceholder that can carry both qubits and
+                # clbits. A plain Barrier cannot carry clbits.
+                placeholder = VerbatimPlaceholder(len(node.qargs), len(node.cargs), label=label)
+                dag.substitute_node(node, placeholder)
             else:
+                barrier = Barrier(len(node.qargs), label=label)
```
laurencap: Does this need to be a Barrier? Why not use `VerbatimPlaceholder` again? If we can, it might simplify the logic below, as well.

qiskit_braket_provider/providers/compilation.py
```diff
@@ -206,6 +209,8 @@ def _compile(
                 )
             )
         ):
+            if has_verbatim_boxes and target and "barrier" not in target.operation_names:
+                target.add_instruction(VerbatimPlaceholder(1, 0), name="barrier")
```
laurencap: Can you explain to me what this line is doing? We have to modify the target?

laurencap: What does `(1, 0)` indicate?

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -793,3 +793,93 @@ def test_verbatim_box_preserves_non_contiguous_qubit_order():
...
+                "optimization_level": 3,
+                "seed_transpiler": 42,
+            },
+            id="basis_gates_opt3",
+        ),
+    ],
+)
+def test_verbatim_box_with_clbits_no_panic(compile_kwargs: dict):
+    """Verbatim BoxOp with clbits does not panic during extract/restore pipeline."""
+    body = QuantumCircuit(2, 2)
```
laurencap: Can we test a Circuit or OpenQASM string with verbatim? My understanding is that is how verbatim would be used by QBP users, since it is more of a Braket construct. If that's right, we should imitate how we think customers would hit the new code.

laurencap (changes requested):

qiskit_braket_provider/providers/compilation.py
aniksd-braket: Yes, we have to modify the target to register this new placeholder instruction with it. The VerbatimPlaceholder(1, 0) is just a dummy instance to register the name "barrier" with the target. The target needs an instruction instance to register, but it doesn't care about the dimensions — it only uses the name for the allowlist check.

qiskit_braket_provider/providers/passes/verbatim_passes.py
aniksd-braket: Oh you're right. I don't need to complicate this

qiskit_braket_provider/providers/passes/verbatim_passes.py
aniksd-braket: That's a good idea, can make labels required as we always pass them anyway

test/unit_tests/test_adapter_to_braket_verbatim.py
aniksd-braket: Sounds good. I had it written in Qiskit here as this was the translated circuit that was causing the bug. But it would make more sense to have the real entry-points instead

qiskit_braket_provider/providers/passes/verbatim_passes.py
```diff
@@ -19,8 +19,23 @@ def _is_verbatim_label(label: str | None, base: str) -> bool:
...
+    Uses the name ``'barrier'`` so the transpiler's basis translator automatically
+    allows it without needing explicit registration in the target or basis_gates.
+    Unlike a real ``Barrier``, this instruction can carry classical bits.
+    """
+
+    _directive = True
+
+    def __init__(self, num_qubits: int, num_clbits: int, label: str):
+        super().__init__("barrier", num_qubits, num_clbits, [])
+        self.label = label
```
nilsquet:
```suggestion
        super().__init__("barrier", num_qubits, num_clbits, [], label=label)
```

qiskit_braket_provider/providers/compilation.py
```diff
@@ -206,6 +209,8 @@ def _compile(
                 )
             )
         ):
+            if has_verbatim_boxes and target and "barrier" not in target.operation_names:
+                target.add_instruction(VerbatimPlaceholder(1, 0, label="verbatim"), name="barrier")
```
nilsquet: The class docstring says "automatically allows it without needing explicit registration in the target or basis_gates" — this line contradicts that. Dropping it (+ the import) passes 471/471 tests. 

See this example:
  ```python
  from qiskit import QuantumCircuit, generate_preset_pass_manager
  from qiskit.circuit.library import HGate, Measure
  from qiskit_braket_provider.providers.passes.verbatim_passes import VerbatimPlaceholder
  
  t = Target(num_qubits=2)
  for g in (HGate(), Measure()):
      t.add_instruction(g)
  assert "barrier" not in t.operation_names
  
  qc = QuantumCircuit(2, 2)
  qc.h(0)
  qc.append(VerbatimPlaceholder(2, 2, label="verbatim__0"), [0, 1], [0, 1])
  qc.measure(0, 0)
  qc.measure(1, 1)
  
  out = generate_preset_pass_manager(3, target=t, layout_method="trivial", routing_method="none").run(qc)
  print([type(i.operation).__name__ for i in out.data])
  # ['_SingletonHGate', 'VerbatimPlaceholder', '_SingletonMeasure', '_SingletonMeasure']
```

Also, add_instruction mutates the caller's target in place, leaving "barrier" (a VerbatimPlaceholder with label="verbatim") on it after the first call.

File: qiskit_braket_provider/providers/passes/verbatim_passes.py
Comment (nilsquet) [inline qiskit_braket_provider/providers/passes/verbatim_passes.py]: Nit: after the refactor, three leftover uses of "barrier" where "placeholder" is now the accurate term — _is_verbatim_label's docstring ("verbatim barrier label"), RestoreVerbatimBoxes's verbatim_box_name arg description ("placeholder barriers"), and the leftover-boxes  RuntimeError message ("no matching barriers found ... Their barriers may have been removed"). Worth updating for consistency with the new VerbatimPlaceholder-based implementation.

File: test/unit_tests/test_adapter_to_braket_verbatim.py [hunk 7]
```diff
@@ -793,3 +796,112 @@ def test_verbatim_box_preserves_non_contiguous_qubit_order():
     out = _compile(qc, basis_gates={"cx"}).circuits[0]
     cx = next(i for i in out.data if i.operation.name == "cx")
     assert [out.find_bit(q).index for q in cx.qubits] == [2, 0]
+
+
+def _build_target_2q() -> Target:
```
nilsquet: helper functions should go to the top

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -793,3 +796,112 @@ def test_verbatim_box_preserves_non_contiguous_qubit_order():
...
+    source = """\
+OPENQASM 3.0;
+bit[2] c;
+qubit[2] q;
+h q[0];
+#pragma braket verbatim
+box { rx(0.5) q[0]; cz q[0], q[1]; }
+c[0] = measure q[0];
+c[1] = measure q[1];
+"""
```
yitchen-tim, line 842: How much does this test case add to the existing ones? The box does not really se cbit? (maybe what matters is the program include c bits? But we already have test case for that too)

aniksd-braket: The existing tests didn't catch this bug because they either:
  
  1. Used basis_gates= at optimization_level=0 (which happened not to trigger the panic)
  2. Used box bodies with fewer qubits/clbits (e.g., QuantumCircuit(1, 1)) or no clbits at all, where the wiring happened to work
  3. Didn't use the target= path at all

test/unit_tests/test_adapter_to_braket_verbatim.py
```diff
@@ -40,6 +43,19 @@ def _make_box_circuit(num_qubits: int, gates: list[tuple[str, list[int]]]) -> Qu
     return qc
 
 
+def _build_target_2q() -> Target:
```
laurencap, line 46: Just to double check, do you still need this? I had thought it had been added to help test "target.add_instruction(VerbatimPlaceholder)", but those lines have been removed.

aniksd-braket: Yup we do still need it, to test compilation when a target is provided.

laurencap (approved):

PR #377, opened 2026-07-28
feature: warn when qubit mapping may be lost with add_measurements=False

### LINKED ISSUE #287: qubit register information lost without measurements in to_braket
`to_braket` function, without measurements, carries an implicit logical to physical mapping, which generally is lost upon translation. 

MWE: 
```
from qiskit_braket_provider import to_braket

test = Circuit().h(0).h(0).rz(1,0.1).rz(1,-0.1).cnot(2,3).cnot(3,4).cnot(4,5)
print(test)
cmap = [[0,1],[1,0],[1,2],[2,1],[2,3],[3,2],[3,4],[4,3],[4,5],[5,4]]
test = to_braket(test, add_measurements=False, optimization_level=2, coupling_map = cmap)
print(test)
```
Circuit is remapped to a zero-indexed circuit, i.e. from 
```
T  : │     0      │      1      │  2  │
         ┌───┐         ┌───┐           
q0 : ────┤ H ├─────────┤ H ├───────────
         └───┘         └───┘           
      ┌──────────┐ ┌───────────┐       
q1 : ─┤ Rz(0.10) ├─┤ Rz(-0.10) ├───────
      └──────────┘ └───────────┘       
                                       
q2 : ──────●───────────────────────────
           │                           
         ┌─┴─┐                         
q3 : ────┤ X ├───────────●─────────────
         └───┘           │             
                       ┌─┴─┐           
q4 : ──────────────────┤ X ├───────●───
                       └───┘       │   
                                 ┌─┴─┐ 
q5 : ────────────────────────────┤ X ├─
                                 └───┘ 
T  : │     0      │      1      │  2  │
```
to 
```T  : │  0  │  1  │  2  │
                        
q0 : ───●───────────────
        │               
      ┌─┴─┐             
q1 : ─┤ X ├───●─────────
      └───┘   │         
            ┌─┴─┐       
q2 : ───────┤ X ├───●───
            └───┘   │   
                  ┌─┴─┐ 
q3 : ─────────────┤ X ├─
                  └───┘ 
T  : │  0  │  1  │  2  │
```

(1) it is unclear whether or not Q0 and Q1 should have been lost, and (2) it is hard to imply the original logical to physical mapping without measure instructions.

Adithyaphani:
Issue #287

**Description of changes:**

When `to_braket` is called with `optimization_level > 0` and a `coupling_map`
(or `target`) and `add_measurements=False`, the Qiskit transpiler remaps virtual
qubits to physical device qubits. If gates cancel during optimization (e.g. H·H
or Rz(x)·Rz(-x)), those qubits disappear from the output entirely. The surviving
gates land on physical qubits chosen by the router — which may not match the
original qubit indices the caller expected. Without measurements there is nothing
in the Braket circuit output to indicate this remapping occurred, so the mapping
is silently lost.

A `UserWarning` is now issued in `to_braket` when `add_measurements=False` and
the transpiler produced a non-trivial qubit permutation (layout ≠ identity). The
message directs users toward `add_measurements=True` or `optimization_level=0`
when qubit correspondence must be preserved.

**Testing done:**

4 new tests in `test/unit_tests/test_adapter.py`:

- `test_qubit_mapping_warning_fires_when_optimization_remaps_qubits` — warning fires with `optimization_level=2` + `coupling_map` + `add_measurements=False`
- `test_no_qubit_mapping_warning_with_measurements` — no warning with `add_measurements=True`
- `test_no_qubit_mapping_warning_without_coupling_map` — no warning without `coupling_map`
- `test_no_qubit_mapping_warning_with_trivial_layout` — no warning with `optimization_level=0`

Full suite: 353 passed, 4 skipped, 0 failures.

---
*Note: AI assistance (Claude) was used during diagnosis.*

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/adapter.py
+++ b/qiskit_braket_provider/providers/adapter.py
@@ -119,6 +119,14 @@ def _check_positional(pos: _T, kw: _T, name: str) -> _T:
     return pos
 
 
+def _has_nontrivial_layout(circuit: QuantumCircuit) -> bool:
+    """Return True if the transpiler assigned a non-identity qubit permutation."""
+    if circuit.layout is None:
+        return False
+    layout = circuit.layout.initial_index_layout(filter_ancillas=False)
+    return layout is not None and layout != list(range(len(layout)))
+
+
 @overload
 def to_braket(
     circuits: _Translatable = ...,
@@ -270,6 +278,18 @@ def to_braket(
         )
         for circ in result.circuits
     ]
+    if not add_measurements and any(_has_nontrivial_layout(c) for c in result.circuits):
+        warnings.warn(
+            "Transpilation with optimization_level > 0 and a coupling_map or target "
+            "may remap qubits. When add_measurements=False the logical-to-physical "
+            "qubit mapping is not reflected by measurement assignments and can be "
+            "hard to recover from the output circuit. "
+            "Consider using add_measurements=True, optimization_level=0, or "
+            "removing the coupling_map (which implicitly assumes a virtual-to-physical "
+            "mapping) if qubit correspondence must be preserved.",
+            UserWarning,
+            stacklevel=2,
+        )
     return translated[0] if single_instance else translated
 
 
--- a/test/unit_tests/test_adapter.py
+++ b/test/unit_tests/test_adapter.py
@@ -1,6 +1,7 @@
 """Tests for Qiskit to Braket adapter."""
 
 import copy
+import warnings
 from unittest import TestCase
 from unittest.mock import MagicMock, Mock, patch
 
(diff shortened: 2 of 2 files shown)


Adithyaphani:
@sesmart looking forward to your review on this and happy to make further changes if needed.

Adithyaphani:
@sesmart looking forward to your review and happy to make further changes if needed.

Adithyaphani:
@sesmart Fixed the two CI failures:

- **mypy**: added an explicit `return False` at the end of `_has_nontrivial_layout`
  to give mypy a clear return path through all branches.
- **ruff format**: reformatted the affected lines to match the project style.

Looking forward to your review and happy to make further changes if needed.

sesmart:
Seems to still have some issues? @Adithyaphani

Adithyaphani:
@sesmart , Fixed the remaining CI failures:

- **mypy + ruff**: moved `_has_nontrivial_layout` to module level (before the
  `@overload` decorators) so mypy sees a proper return statement and ruff
  formats it correctly. The earlier nested placement was between the overloads
  and the implementation, which broke Python's overload resolution entirely.

sesmart:
Hi @Adithyaphani - I would just run `tox` locally to confirm that it passes - seems like a minor formatting issue.

Adithyaphani:
@sesmart , Ran tox locally and fixed the two remaining issues:

- **ruff**: renamed `BLE001` to `blind-except` to match the rule name format
  the newer ruff version requires, and added `pragma: no cover` on the
  defensive except branch since it guards against future Qiskit API changes
  that can't be triggered in tests.
- **coverage**: the uncovered lines in `_has_nontrivial_layout` are now
  excluded via `pragma: no cover`.

Adithyaphani:
@sesmart  looking forward to your review and happy to make further changes.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -119,6 +119,22 @@ def _check_positional(pos: _T, kw: _T, name: str) -> _T:
     return pos
 
 
+def _has_nontrivial_layout(circuit: QuantumCircuit) -> bool:
+    """Return True if the transpiler assigned a non-identity qubit permutation.
+
+    A non-trivial layout means the transpiler placed virtual qubits onto
```
sesmart: Detailed docstring can be removed for private method.

qiskit_braket_provider/providers/adapter.py
```diff
@@ -119,6 +119,22 @@ def _check_positional(pos: _T, kw: _T, name: str) -> _T:
...
+    physical qubits in a permuted order.  When ``add_measurements=False``
+    this remapping is invisible in the Braket output circuit.
+    """
+    if circuit.layout is None:
+        return False
+    try:
+        layout = circuit.layout.initial_index_layout(filter_ancillas=False)
+        return layout is not None and layout != list(range(len(layout)))
+    except Exception:  # ruff: ignore[blind-except]
```
sesmart: Why do we need this exception? The check is already checking for `TranspilerLayout` or `None`....and I don't see a `raise` in any of the sub functions, so I don't think we need to guard it, unless I am missing something?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -119,6 +119,22 @@ def _check_positional(pos: _T, kw: _T, name: str) -> _T:
...
+
+    A non-trivial layout means the transpiler placed virtual qubits onto
+    physical qubits in a permuted order.  When ``add_measurements=False``
+    this remapping is invisible in the Braket output circuit.
+    """
+    if circuit.layout is None:
+        return False
+    try:
+        layout = circuit.layout.initial_index_layout(filter_ancillas=False)
+        return layout is not None and layout != list(range(len(layout)))
```
sesmart: So this just checks if it is the same trivial layout? What if have a (0,1,3) Braket Circuit, which i contiguously mapped to (0,1,2), and then matches? Is there any way to return then?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -270,6 +286,17 @@ def to_braket(
         )
         for circ in result.circuits
     ]
+    if not add_measurements and any(_has_nontrivial_layout(c) for c in result.circuits):
```
sesmart, line 281: I do like this warning a lot, though should it be in _compile?

qiskit_braket_provider/providers/adapter.py
```diff
@@ -270,6 +286,17 @@ def to_braket(
...
+        warnings.warn(
+            "Transpilation with optimization_level > 0 and a coupling_map or target "
+            "may remap qubits. When add_measurements=False the logical-to-physical "
+            "qubit mapping is not reflected by measurement assignments and can be "
+            "hard to recover from the output circuit. "
+            "Consider using add_measurements=True or optimization_level=0 if qubit "
```
sesmart: Could add, "or removing the coupling map, which implicitly assumes a virtual to physical mapping." 

We also could recommend adding qubit labels, if that is sufficient to help with losing unused qubits.

test/unit_tests/test_adapter.py
```diff
@@ -1628,6 +1628,113 @@ def test_oq3_parametric_function_direct_composite(self):
...
 
+    def test_qubit_mapping_warning_fires_when_optimization_remaps_qubits(self):
+        """Tests that a UserWarning is issued when optimization_level > 0 and a
+        coupling_map cause qubit remapping with add_measurements=False.
+
+        Without measurements there is no way to recover the logical-to-physical
+        qubit mapping from the Braket circuit output alone. The warning alerts
+        callers before this information is silently lost. Resolves issue #287.
+        """
+        import warnings
```
sesmart: No local imports

test/unit_tests/test_adapter.py
```diff
@@ -1628,6 +1628,113 @@ def test_oq3_parametric_function_direct_composite(self):
...
+            msg="Expected exactly one UserWarning about qubit remapping when "
+            "optimization_level > 0 and add_measurements=False.",
+        )
+        self.assertIn("add_measurements=True", str(mapping_warnings[0].message))
+
+    def test_no_qubit_mapping_warning_with_measurements(self):
+        """No qubit-remapping warning when add_measurements=True.
+        Resolves issue #287.
+        """
+        import warnings
```
sesmart: Same

sesmart (commented):
I like the warning, let me know if you have any questions! Also I realized we generally lose ghost qubits, i.e. if Braket specifies Q0, Q1, Q3 - this will always to remapped to a contiguous set. Maybe we could harness qubit_labels as well within this, but will check with the team if this is intended behavior.

qiskit_braket_provider/providers/adapter.py
Adithyaphani: Done — shortened to one line.

qiskit_braket_provider/providers/adapter.py
Adithyaphani: Good catch — removed. The None check already guards the only failure mode and initial_index_layout doesn't raise.

qiskit_braket_provider/providers/adapter.py
Adithyaphani: Good point — if Braket specifies non-contiguous qubits (e.g. q0,q1,q3) and the transpiler maps them to contiguous physical qubits (0,1,2), the identity check won't catch it. Agree this is worth a follow-up, possibly via qubit_labels. Happy to track it as a separate issue.

qiskit_braket_provider/providers/adapter.py
Adithyaphani: _compile doesn't receive add_measurements so the check would need to be split across both functions. Happy to move it if you have a preference, otherwise keeping it in to_braket where the full context is available seems cleaner.

qiskit_braket_provider/providers/adapter.py
Adithyaphani: Done — added "or removing the coupling_map".

test/unit_tests/test_adapter.py
Adithyaphani: Done — moved import warnings to module level.

test/unit_tests/test_adapter.py
Adithyaphani: Done — moved to module level.

Adithyaphani:
@sesmart ,Addressed all the review comments — shortened the docstring, dropped the
try/except (the None check is sufficient), updated the warning to mention
removing the coupling_map, and moved import warnings to module level.

On the (0,1,3) → (0,1,2) edge case — good catch, happy to track that as a
follow-up once this lands.

On moving the warning to _compile — since _compile doesn't receive
add_measurements, the check would need to be split. Happy to move it if you
prefer, but keeping it in to_braket where the full context is available felt
cleaner. Let me know!

qiskit_braket_provider/providers/adapter.py
sesmart: Ah, good point. You are 100% right.

qiskit_braket_provider/providers/adapter.py
sesmart: Yep, agree a separate issue is likely appropriate.

sesmart (approved):

sesmart:
Am looking into the contiguous bit, I agree it can be tracked as a separate issue. LGTM now, thanks!

Adithyaphani:
@sesmart sent you a LinkedIn request many weeks ago , when you find free time please accept it.

PR #389, opened 2026-08-24
feat: add OQ3 post-processing utilities

aniksd-braket:
Adds string-manipulation helpers that transform the raw output of qasm3.dumps() into Braket-compatible OpenQASM 3.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/qiskit_braket_provider/providers/braket_annotations.py
+++ b/qiskit_braket_provider/providers/braket_annotations.py
@@ -1,0 +1,63 @@
+"""Braket-verbatim annotation and box-op infrastructure for OpenQASM 3 emission.
+
+This module hosts the Qiskit :class:`~qiskit.circuit.annotation.Annotation` and
+:class:`~qiskit.circuit.BoxOp` machinery used to mark verbatim blocks in Qiskit
+circuits so that ``qasm3.dumps`` emits Braket's ``#pragma braket verbatim``
+directive.
+"""
+
+import re
+
+from qiskit.circuit import BoxOp, QuantumCircuit
+from qiskit.circuit.annotation import Annotation, OpenQASM3Serializer
+
+from qiskit_braket_provider.providers.gate_mappings import (
+    _BRAKET_VERBATIM_BOX_NAME,
+    _BRAKET_VERBATIM_PRAGMA_PAYLOAD,
+)
+
+_BRAKET_VERBATIM_ANNOTATION_LINE = re.compile(r"^@braket_verbatim\s+pragma\s+braket\s+verbatim\s*$")
+
+
+class BraketVerbatim(Annotation):
+    """Marks a ``BoxOp`` as a Braket verbatim block for OpenQASM 3 emission."""
+
+    namespace = "braket_verbatim"
+
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, BraketVerbatim)
+
+    def __hash__(self) -> int:
+        return hash(BraketVerbatim)
+
+
+class BraketVerbatimSerializer(OpenQASM3Serializer):
+    """Serializes :class:`BraketVerbatim` as ``pragma braket verbatim``."""
+
+    def dump(self, annotation: Annotation) -> str:
+        if isinstance(annotation, BraketVerbatim):
+            return _BRAKET_VERBATIM_PRAGMA_PAYLOAD
+        raise NotImplementedError  # pragma: no cover
+
+    def load(self, _namespace: str, _payload: str) -> Annotation:
+        raise NotImplementedError  # pragma: no cover
+
+
+_BRAKET_ANNOTATION_HANDLERS: dict[str, OpenQASM3Serializer] = {
+    BraketVerbatim.namespace: BraketVerbatimSerializer(),
+}
+
+
+class BraketVerbatimBox(BoxOp):
+    """A ``BoxOp`` pre-configured as a Braket verbatim block.
+
+    Attaches :class:`BraketVerbatim` and sets ``label=_BRAKET_VERBATIM_BOX_NAME``
+    (overridable via ``label``) so ``qasm3.dumps`` emits the verbatim directive.
+    """
+
+    def __init__(self, body: QuantumCircuit, label: str = _BRAKET_VERBATIM_BOX_NAME) -> None:
+        super().__init__(
+            body,
+            label=label,
+            annotations=[BraketVerbatim()],
+        )
--- a/qiskit_braket_provider/providers/gate_mappings.py
+++ b/qiskit_braket_provider/providers/gate_mappings.py
@@ -215,6 +215,10 @@ _BRAKET_GATE_NAME_TO_QISKIT_GATE: dict[str, QiskitInstruction | None] = {
 
 _BRAKET_VERBATIM_BOX_NAME = "verbatim"
 
+_BRAKET_VERBATIM_PRAGMA_PAYLOAD = "pragma braket verbatim"
+
+_BRAKET_VERBATIM_PRAGMA_LINE = f"#{_BRAKET_VERBATIM_PRAGMA_PAYLOAD}"
+
 _OUTPUT_VARIABLES_KEY = "braket_output_variables"
 
 _EPS = 1e-10  # global variable used to chop very small numbers to zero
@@ -291,3 +295,135 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
         matrix.reshape([2] * n_q * 2),
         list(range(n_q))[::-1] + list(range(n_q, 2 * n_q))[::-1],
     ).reshape((2**n_q, 2**n_q))
+
+
+class BraketPhaseShiftGate(qiskit_gates.PhaseGate):
+    """Qiskit ``PhaseGate`` rendered as Braket's ``phaseshift`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "phaseshift"
+
+
+class BraketCNotGate(qiskit_gates.CXGate):
+    """Qiskit ``CXGate`` rendered as Braket's ``cnot`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "cnot"
+
+
+class BraketTiGate(qiskit_gates.TdgGate):
+    """Qiskit ``TdgGate`` rendered as Braket's ``ti`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "ti"
+
+
+class BraketSiGate(qiskit_gates.SdgGate):
+    """Qiskit ``SdgGate`` rendered as Braket's ``si`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "si"
+
+
+class BraketVGate(qiskit_gates.SXGate):
+    """Qiskit ``SXGate`` rendered as Braket's ``v`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "v"
+
+
+class BraketViGate(qiskit_gates.SXdgGate):
+    """Qiskit ``SXdgGate`` rendered as Braket's ``vi`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "vi"
+
+
+class BraketXXGate(qiskit_gates.RXXGate):
+    """Qiskit ``RXXGate`` rendered as Braket's ``xx`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "xx"
+
+
+class BraketYYGate(qiskit_gates.RYYGate):
+    """Qiskit ``RYYGate`` rendered as Braket's ``yy`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "yy"
+
+
+class BraketZZGate(qiskit_gates.RZZGate):
+    """Qiskit ``RZZGate`` rendered as Braket's ``zz`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "zz"
+
+
+class BraketIGate(qiskit_gates.IGate):
+    """Qiskit ``IGate`` rendered as Braket's ``i`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "i"
+
+
+class BraketCCNotGate(qiskit_gates.CCXGate):
+    """Qiskit ``CCXGate`` rendered as Braket's ``ccnot`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "ccnot"
+
+
+class BraketCPhaseShiftGate(qiskit_gates.CPhaseGate):
+    """Qiskit ``CPhaseGate`` rendered as Braket's ``cphaseshift`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "cphaseshift"
+
+
+class BraketPRxGate(qiskit_gates.RGate):
+    """Qiskit ``RGate`` rendered as Braket's ``prx`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "prx"
+
+
+class BraketGPhaseGate(qiskit_gates.GlobalPhaseGate):
+    """Qiskit ``GlobalPhaseGate`` rendered as Braket's ``gphase`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "gphase"
+
+
+_QISKIT_TO_BRAKET_SHIM: dict[type, type] = {
+    qiskit_gates.PhaseGate: BraketPhaseShiftGate,
+    qiskit_gates.CXGate: BraketCNotGate,
+    qiskit_gates.TdgGate: BraketTiGate,
+    qiskit_gates.SdgGate: BraketSiGate,
+    qiskit_gates.SXGate: BraketVGate,
+    qiskit_gates.SXdgGate: BraketViGate,
+    qiskit_gates.RXXGate: BraketXXGate,
+    qiskit_gates.RYYGate: BraketYYGate,
+    qiskit_gates.RZZGate: BraketZZGate,
+    qiskit_gates.IGate: BraketIGate,
+    qiskit_gates.CCXGate: BraketCCNotGate,
+    qiskit_gates.CPhaseGate: BraketCPhaseShiftGate,
+    qiskit_gates.RGate: BraketPRxGate,
+    qiskit_gates.GlobalPhaseGate: BraketGPhaseGate,
+}
+
+_SHIM_CLASSES: frozenset[type] = frozenset(_QISKIT_TO_BRAKET_SHIM.values())
--- a/qiskit_braket_provider/providers/passes/braket_formatting_passes.py
+++ b/qiskit_braket_provider/providers/passes/braket_formatting_passes.py
@@ -1,24 +1,27 @@
-"""Transpiler passes for preparing circuits for Braket-compatible serialization."""
+"""Transpiler passes and string post-processing helpers for Braket-compatible OQ3 emission."""
 
-from qiskit.circuit import BoxOp, ClassicalRegister, Measure, QuantumCircuit
+import re
+from collections.abc import Iterable, Sequence
+
+from qiskit.circuit import ClassicalRegister, Instruction, Measure, QuantumCircuit
 from qiskit.converters import circuit_to_dag, dag_to_circuit
 from qiskit.dagcircuit import DAGCircuit
 from qiskit.transpiler.basepasses import TransformationPass
 
+from qiskit_braket_provider.providers.braket_annotations import (
+    _BRAKET_VERBATIM_ANNOTATION_LINE,
+    BraketVerbatimBox,
+)
 from qiskit_braket_provider.providers.gate_mappings import (
-    _BRAKET_VERBATIM_BOX_NAME,
+    _BRAKET_VERBATIM_PRAGMA_LINE,
     _OUTPUT_VARIABLES_KEY,
+    _QISKIT_TO_BRAKET_SHIM,
+    _SHIM_CLASSES,
 )
 
+_QUBIT_DECL_RE = re.compile(r"^qubit\[(\d+)\]\s+(\w+);\n?", re.MULTILINE)
 
-def _plain_register_name(taken: set) -> str:
-    """Return ``"b"``, or the first ``"b<i>"`` not shadowed by an output variable name."""
-    if "b" not in taken:
-        return "b"
-    i = 0
-    while f"b{i}" in taken:
-        i += 1
-    return f"b{i}"
+_PHYSICAL_QUBIT_RE = re.compile(r"\$(\d+)")
 
 
 class ConsolidateClbits(TransformationPass):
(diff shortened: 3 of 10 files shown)


qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    """Post-process qasm3.dumps() output for Braket compatibility.
+
+    Performs the following transformations:
+    1. Renames Qiskit gate names to Braket-compatible OQ3 gate names (when needed)
+    2. Remaps virtual qubit references to physical qubit labels
+    3. Inserts ``#pragma braket verbatim`` before any ``box`` statements
+    4. Replaces ``float[64]`` with ``float`` for parameter declarations
+    5. Restores ``output`` keyword on classical register declarations whose name
+       is listed in ``output_names``
+    6. Strips indentation and removes empty lines (Braket convention)
```
nilsquet: Below, there are only 3 function calls. Some of these steps are in the respectively called methods. In case the methods are modified, it's likely this docstring will be overlooked and not updated. Therefore, please only document the things that this method is doing.

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        if stripped.startswith(("OPENQASM", "//")):
+            result.append(line)
+            continue
+        result.append(_GATE_RENAME_PATTERN.sub(_gate_replacer, line))
+    return "\n".join(result)
+
+
+def _gate_replacer(match: re.Match) -> str:
+    """Regex replacer callback for gate renaming."""
+    return _QISKIT_TO_BRAKET_OQ3_NAMES[match.group(1)]
```
nilsquet: Can we do this as a regular `TransformationPass` instead? We should avoid the regex postprocessing as much as possible.

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Handles two cases produced by ``qasm3.dumps()``:
+
+    1. **Layout-aware output** (``$N`` notation, no ``qubit[N] q;`` declaration):
+       when the circuit's ``layout`` property is set, Qiskit emits physical
+       qubit references directly. Remap ``$N`` → ``$qubit_labels[N]``.
+
+    2. **Virtual register output** (``qubit[N] q; ... q[i]``): when no layout
+       is attached, Qiskit falls back to virtual register syntax. Remove the
+       ``qubit[N] q;`` declaration and remap ``q[i]`` → ``$qubit_labels[i]``.
```
nilsquet: I don't understand this case. If there is not layout attached, why would be effectively apply a trivial layout?

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    - Inserts ``#pragma braket verbatim`` before ``box`` statements
+    - Removes space in ``box {`` → ``box{``
+    - Restores ``output`` keyword on registers listed in ``output_names``
+    - Removes empty lines
+    """
+    output_set = set(output_names)
+    lines: list[str] = []
+    for raw_line in oq3_source.split("\n"):
+        line = raw_line.lstrip().replace("float[64]", "float")
+        if line == "box {":
```
nilsquet:
```suggestion
        if re.match(r"^box\s*\{?\s*$", line):
```
This way, more cases (`box{`, `box  {`, and trailing whitespace) are handled in case Qiskit's dumps would change.

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    oq3_source = _remap_qubits(oq3_source, qubit_labels)
+    return _normalize_formatting(oq3_source, output_names)
+
+
+def _rename_gates(oq3_source: str) -> str:
+    """Replace Qiskit gate names with Braket-compatible names."""
+    result = []
+    for line in oq3_source.split("\n"):
+        stripped = line.lstrip()
+        if stripped.startswith(("OPENQASM", "//")):
```
nilsquet: Why would we strip comments?

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    - Removes space in ``box {`` → ``box{``
+    - Restores ``output`` keyword on registers listed in ``output_names``
+    - Removes empty lines
+    """
+    output_set = set(output_names)
+    lines: list[str] = []
+    for raw_line in oq3_source.split("\n"):
+        line = raw_line.lstrip().replace("float[64]", "float")
+        if line == "box {":
+            lines.extend(("#pragma braket verbatim", "box{"))
```
nilsquet: This now unconditionally translates every `box` to a `pragma braket verbatim box`, which is wrong.

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            continue
+        lines.append(_restore_output_declaration(line, output_set))
+    return "\n".join(line for line in lines if line != "")
+
+
+def _restore_output_declaration(line: str, output_names: set) -> str:
+    """Prefix ``output`` onto a bit declaration ``qasm3.dumps`` rendered as plain.
```
nilsquet: This is hard to understand.
```suggestion
    """Re-add the output keyword to bit declarations whose name is in output_names.
```

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    for raw_line in oq3_source.split("\n"):
+        line = raw_line.lstrip().replace("float[64]", "float")
+        if line == "box {":
+            lines.extend(("#pragma braket verbatim", "box{"))
+            continue
+        lines.append(_restore_output_declaration(line, output_set))
+    return "\n".join(line for line in lines if line != "")
+
+
+def _restore_output_declaration(line: str, output_names: set) -> str:
```
nilsquet:
```suggestion
def _restore_output_declaration(line: str, output_names: set[str]) -> str:
```

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -75,6 +76,22 @@
     "kraus": "kraus",
 }
 
+_QISKIT_TO_BRAKET_OQ3_NAMES: dict[str, str] = {
```
nilsquet:
```suggestion
_QISKIT_TO_BRAKET_NAMES: dict[str, str] = {
```
Since it just reserves `_BRAKET_TO_QISKIT_NAMES` without `OQ3` being in the name.

qiskit_braket_provider/providers/oq3_utils.py
```diff
@@ -0,0 +1,131 @@
+"""Post-processing helpers for OpenQASM 3 emission targeting Amazon Braket.
```
nilsquet: Why are we not having these passes in `braket_formatting_passes.py`?

test/unit_tests/test_oq3_utils.py
```diff
@@ -0,0 +1,157 @@
+"""Tests for the OQ3 post-processing utilities."""
```
nilsquet: Same here, I think this should go to `test_braket_formatting_passes.py`.

test/unit_tests/test_oq3_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+import pytest
+
```
nilsquet: I'd like to see some integration tests that actually use Qiskit's `dumps` as input.

nilsquet (dismissed):

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -291,3 +291,133 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
         matrix.reshape([2] * n_q * 2),
         list(range(n_q))[::-1] + list(range(n_q, 2 * n_q))[::-1],
     ).reshape((2**n_q, 2**n_q))
+
+
+class BraketCnotGate(qiskit_gates.CXGate):
```
nilsquet: The 14 shim classes each encode one fact that _BRAKET_TO_QISKIT_NAMES already states ("cnot": "cx", "ti": "tdg", ...), so the same mapping now lives in two places. Nothing catches drift between them: add a gate to _BRAKET_TO_QISKIT_NAMES without writing a matching shim and there's no error and no failing test — the Qiskit gate name just ends up in the emitted program.
  
  Suggest deriving the shims from the existing table:
  ```python
  def _make_shim(base: type, braket_name: str) -> type:
      def __init__(self, *args, **kwargs) -> None:
          base.__init__(self, *args, **kwargs)
          self.name = braket_name
      return type(f"Braket{base.__name__}", (base,), {"__init__": __init__})
  
  
  _QISKIT_TO_BRAKET_SHIM: dict[type, type] = {
      (base := get_standard_gate_name_mapping()[qiskit_name].base_class): _make_shim(
          base, braket_name
      )
      for braket_name, qiskit_name in _BRAKET_TO_QISKIT_NAMES.items()
      if braket_name != qiskit_name
  }
  ```
  This produces the same 14 entries, and each generated class matches the hand-written one on name and num_qubits including the parametrized gates. RenameGates, _shim_class_for and _construct_shim are unaffected, and the emitted OQ3 is unchanged.
  
 However, the cost is that isinstance(op, BraketCnotGate) in the tests becomes isinstance(op, _QISKIT_TO_BRAKET_SHIM[CXGate]), which reads worse.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +131,175 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        for node in dag.op_nodes():
+            shim_cls = _shim_class_for(node.op)
+            if shim_cls is not None:
+                dag.substitute_node(node, _construct_shim(node.op, shim_cls))
+            elif getattr(node.op, "blocks", None):
+                dag.substitute_node(node, _rename_gates_in_block_op(node.op))
+        return dag
+
+
+def _post_process_oq3(
```
nilsquet: The name is generic, should we rather call it `remap_qubits_and_normalize_formatting`? Alternatively, we could just call the two passes individually where we would use them.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +131,175 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            ``metadata["braket_output_variables"]``.
+
+    Returns:
+        The post-processed OpenQASM 3 string.
+    """
+    oq3_source = _remap_qubits(oq3_source, qubit_labels)
+    return _normalize_formatting(oq3_source, output_names)
+
+
+def _remap_qubits(oq3_source: str, qubit_labels: Sequence[int] | None) -> str:
```
nilsquet, line 216: _remap_qubits only handles the first qubit[N] ...; declaration, so multi-register circuits break. This is reachable: _compile preserves the user's registers whenever it doesn't run a full transpile (including verbatim=True) and attaches no .layout, so a circuit built from two QuantumRegisters lands here with two declarations.
  
  The length check compares against the first register only, so the reported count is wrong:
  ```
  # OPENQASM 3.0;  qubit[2] a;  qubit[2] bq;  h a[0];  cx a[0], bq[1];
  _remap_qubits(src, [10, 11, 12, 13])
  # ValueError: qubit_labels length (4) does not match the circuit's qubit count (2).
  ```
  And if the label count happens to match the first register, it succeeds and emits an invalid program — the second declaration survives and its references stay virtual alongside physical ones:
  ```
  _remap_qubits(src, [10, 11])
  # OPENQASM 3.0;
  # qubit[2] bq;
  # h $10;
  # cx $10, bq[1];
  ```
  Separately, the layout-aware branch has no bounds check, so an out-of-range $N gives a bare IndexError: list index out of range instead of the ValueError the other branch raises. Both paths are reachable from compile_to_oq3, so these are user-facing errors.
  
  Raises: also documents only the virtual-register case.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +131,175 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+      to Braket's canonical ``#pragma braket verbatim`` directive.
+    - Restores ``output`` keyword on registers listed in ``output_names``
+    - Removes empty lines
+    """
+    output_set = set(output_names)
+    lines: list[str] = []
+    for raw_line in oq3_source.split("\n"):
+        line = raw_line.lstrip().replace("float[64]", "float")
+        if _BRAKET_VERBATIM_ANNOTATION_LINE.match(line):
+            lines.append("#pragma braket verbatim")
```
nilsquet: This should be a constant, similar to `_BRAKET_VERBATIM_BOX_NAME`.

qiskit_braket_provider/providers/braket_annotations.py
```diff
@@ -0,0 +1,54 @@
...
+
+    namespace = "braket_verbatim"
+
+
+class BraketVerbatimSerializer(OpenQASM3Serializer):
+    """Serializes :class:`BraketVerbatim` as ``pragma braket verbatim``."""
+
+    def dump(self, annotation: Annotation) -> str:
+        if isinstance(annotation, BraketVerbatim):
+            return "pragma braket verbatim"
```
nilsquet: This should be a constant, similar to `_BRAKET_VERBATIM_BOX_NAME`.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +131,175 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        return oq3_source
+
+    return re.sub(r"\$(\d+)", lambda m: f"${qubit_labels[int(m.group(1))]}", oq3_source)
+
+
+def _normalize_formatting(oq3_source: str, output_names: Iterable[str] = ()) -> str:
+    """Normalize OQ3 formatting to match Braket service conventions.
+
+    - Strips leading indentation from all lines
+    - Replaces ``float[64]`` with ``float`` in parameter declarations
```
nilsquet: Current this is checked for each line and not just "in parameter declarations".

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +131,175 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+
+
+def _shim_class_for(op: Instruction) -> type | None:
+    """Return the shim class for ``op``, or ``None`` if no shim applies.
+
+    Walks ``type(op).__mro__`` so ops that are Qiskit-generated singleton
+    subclasses (e.g. ``_SingletonCXGate`` for ``CXGate``) match their mapped
+    parent. Returns ``None`` if ``op`` is already a shim instance.
+    """
+    shim_classes = frozenset(_QISKIT_TO_BRAKET_SHIM.values())
```
nilsquet: Nit: Having this outside the function would be more efficient cause it would only be executed once and not with every `_shim_class_for()` call.

qiskit_braket_provider/providers/braket_annotations.py
```diff
@@ -0,0 +1,54 @@
...
+        return NotImplemented  # pragma: no cover
+
+    def load(self, _namespace: str, _payload: str) -> Annotation:
+        return NotImplemented  # pragma: no cover
+
+
+BRAKET_ANNOTATION_HANDLERS: dict[str, OpenQASM3Serializer] = {
```
nilsquet: Should this be private?

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -291,3 +295,24 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
...
+
+
+def _make_shim(base: type, braket_name: str) -> type:
+    """Return a subclass of ``base`` that renders as ``braket_name`` in OpenQASM 3."""
+
+    def init(self: object, *args: object, **kwargs: object) -> None:
+        base.__init__(self, *args, **kwargs)  # type: ignore[misc]
+        self.name = braket_name  # type: ignore[attr-defined]
+
+    return type(f"Braket{base.__name__}", (base,), {"__init__": init})
```
speller26: Why do we need dynamic class construction here? This makes debugging really difficult.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +133,168 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            new_circuit.append(op, inst.qubits, inst.clbits)
+    return new_circuit
+
+
+def _rename_gates_in_block_op(op: Instruction) -> Instruction:
+    """Return a copy of a ``.blocks``-carrying op with each block's gates renamed."""
+    return op.replace_blocks([_rename_gates_in_circuit(block) for block in op.blocks])
+
+
+class RenameGates(TransformationPass):
```
yitchen-tim: nit: This can be the first thing in the file, with private functions following it.

yitchen-tim: I also see functions like `_remap_qubits` are defined but not called. Maybe they are for passes in PR3. Would be good to follow the same style of top-down order.

qiskit_braket_provider/providers/gate_mappings.py
yitchen-tim: Just to clarify, @aniksd-braket , we have a map of qiskit gate name to braket gate name. So is it correct that this is to fill the gap for mapping between Qiskit class and Braket class? 

If so, while it reduce duplication (ie maintaining two maps, one for name and one for class), we are also running on a assumption that the gate name and class name is the same, which is a brittle assumption. And this is on top of the risks Cody calls out.

test/unit_tests/test_braket_formatting_passes.py
```diff
@@ -306,6 +437,364 @@ def test_wrap_in_verbatim_box_empty_circuit(circuit: QuantumCircuit) -> None:
...
+    qc = QuantumCircuit(2)
+    qc.rxx(1.5, 0, 1)
+    qc.p(0.75, 0)
+    result = PassManager([RenameGates()]).run(qc)
+    ops = [instr.operation for instr in result.data]
+    assert ops[0].params == [1.5]
+    assert ops[1].params == [0.75]
+
+
+def test_rename_gates_covers_every_shim() -> None:
```
yitchen-tim, line 558: Maybe we are waiting for PR3 for end to end tests, but just want to add that it's better to compare the exact OQ3 program

qiskit_braket_provider/providers/gate_mappings.py
aniksd-braket: Just for brevity. I could make explicit classes. Would you rather have that? A previous version of the PR had this implemented that way.

test/unit_tests/test_braket_formatting_passes.py
aniksd-braket: Will keep this in mind for PR3

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -291,3 +295,135 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
...
+
+class BraketPhaseshiftGate(qiskit_gates.PhaseGate):
+    """Qiskit ``PhaseGate`` rendered as Braket's ``phaseshift`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "phaseshift"
+
+
+class BraketCnotGate(qiskit_gates.CXGate):
```
speller26: Nit: `CNot...`

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -291,3 +295,135 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
...
+
+class BraketViGate(qiskit_gates.SXdgGate):
+    """Qiskit ``SXdgGate`` rendered as Braket's ``vi`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "vi"
+
+
+class BraketXxGate(qiskit_gates.RXXGate):
```
speller26: nit: `XX...` basically follow Braket name conventions. No big deal though

qiskit_braket_provider/providers/braket_annotations.py
```diff
@@ -0,0 +1,63 @@
...
+        return hash(BraketVerbatim)
+
+
+class BraketVerbatimSerializer(OpenQASM3Serializer):
+    """Serializes :class:`BraketVerbatim` as ``pragma braket verbatim``."""
+
+    def dump(self, annotation: Annotation) -> str:
+        if isinstance(annotation, BraketVerbatim):
+            return _BRAKET_VERBATIM_PRAGMA_PAYLOAD
+        return NotImplemented  # pragma: no cover
```
speller26:
```suggestion
        raise NotImplementedError  # pragma: no cover
```
Returning `NotImplemented` can cause silent problems

qiskit_braket_provider/providers/braket_annotations.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def load(self, _namespace: str, _payload: str) -> Annotation:
+        return NotImplemented  # pragma: no cover
```
speller26: ditto

qiskit_braket_provider/providers/gate_mappings.py
```diff
@@ -291,3 +295,135 @@ def _reverse_endianness(matrix: np.ndarray) -> np.ndarray:
...
+
+class BraketGphaseGate(qiskit_gates.GlobalPhaseGate):
+    """Qiskit ``GlobalPhaseGate`` rendered as Braket's ``gphase`` in OpenQASM 3."""
+
+    def __init__(self, *args: object, **kwargs: object) -> None:
+        super().__init__(*args, **kwargs)
+        self.name = "gphase"
+
+
+_QISKIT_TO_BRAKET_SHIM: dict[type, type] = {
```
speller26, line 412: No need to do this, but you can actually further parametrize `type` here iirc.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +133,168 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+
+    Raises:
+        ValueError: If the source contains more than one ``qubit[N] ...;``
+            declaration. Braket supports at most one quantum register.
+        ValueError: If ``len(qubit_labels)`` does not match the ``N`` declared
+            in the ``qubit[N] q;`` register (virtual-register case), or if a
+            physical qubit reference ``$N`` in the source is out of range for
+            ``qubit_labels`` (layout-aware case).
+    """
+    decl = re.compile(r"^qubit\[(\d+)\]\s+(\w+);\n?", re.MULTILINE)
```
speller26: Do we need this to be a regex? If so, extract to constant.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +133,168 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            oq3_source = oq3_source.replace(f"{reg_name}[{i}]", f"${label}")
+        return oq3_source
+
+    referenced = [int(m) for m in re.findall(r"\$(\d+)", oq3_source)]
+    if referenced and max(referenced) >= len(qubit_labels):
+        raise ValueError(
+            f"qubit_labels length ({len(qubit_labels)}) is too short for "
+            f"physical qubit reference ${max(referenced)} in the source."
+        )
+    return re.sub(r"\$(\d+)", lambda m: f"${qubit_labels[int(m.group(1))]}", oq3_source)
```
speller26: A comment would help tell the reader what this substitution is doing.

qiskit_braket_provider/providers/passes/braket_formatting_passes.py
```diff
@@ -124,9 +133,168 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+def _shim_class_for(op: Instruction) -> type | None:
+    """Return the shim class for ``op``, or ``None`` if no shim applies.
+
+    Walks ``type(op).__mro__`` so ops that are Qiskit-generated singleton
+    subclasses (e.g. ``_SingletonCXGate`` for ``CXGate``) match their mapped
+    parent. Returns ``None`` if ``op`` is already a shim instance.
+    """
+    for cls in type(op).__mro__:
+        if cls in _SHIM_CLASSES:
+            return None
```
speller26, line 177: Is this branch necessary, or is the following one sufficient?

aniksd-braket: It's necessary — Qiskit's singleton machinery creates a _SingletonBraketCNotGate subclass at instantiation time and inserts it at MRO[0], so checking type(op) up-front
  would miss it. The MRO walk needs the shim-guard to short-circuit before falling through to the shim's Qiskit base class. Expanded the docstring to explain this.

qiskit_braket_provider/providers/gate_mappings.py
yitchen-tim: @speller26 Aniket had an implementation with dynamic type before, and you commented that it would be hard to debug. Are you suggesting a different implementation here?

speller26 (approved):

yitchen-tim (approved):

qiskit_braket_provider/providers/gate_mappings.py
speller26: I like the new explicit implementation much more

speller26: All I mean with this comment is that the type hint can actually be even more detailed, but no big deal.

PR #395, opened 2026-09-07
fix: query queue position from the task, not a reconstructed batch id

MohammedAlkindi:
`BraketQuantumTask.queue_position()` cannot work for a multi-circuit job. It loops over `self._tasks` but returns `AwsQuantumTask(self.task_id()).queue_position()`, and for a batch `task_id()` is every task ARN joined with `;` (`_TASK_ID_DIVIDER` in `braket_backend.py`). That string is not an ARN, so the reconstructed task's `metadata()` call fails against the service.

The fix uses the task being iterated, which already holds a valid ARN:

```diff
-            return AwsQuantumTask(self.task_id()).queue_position()
+            return task.queue_position()
```

Single-task jobs, program sets and `LocalQuantumTask` are unchanged; the loop still returns on the first task.

`test_queue_position` used one task, where the joined id equals the task's own id, so it asserted the reconstruction and passed. It now asserts the task's own method is called, and a new test covers two tasks. Both fail on unmodified `main` (`Expected 'queue_position' to have been called once. Called 0 times`) and pass with the change; 30/30 after.

Not verified against live Braket; I have no AWS credentials here. The wider suite could not run on this machine because a Windows policy blocks the `numba` DLL, so I ran this file with `numba` stubbed for import only.

Drafted with AI assistance. I read the code path and ran the tests before and after myself.

### DIFF
--- a/qiskit_braket_provider/providers/braket_quantum_task.py
+++ b/qiskit_braket_provider/providers/braket_quantum_task.py
@@ -157,7 +157,7 @@ class BraketQuantumTask(JobV1):
                 raise NotImplementedError(
                     "We don't provide queue information for the LocalQuantumTask."
                 )
-            return AwsQuantumTask(self.task_id()).queue_position()
+            return task.queue_position()
         return None
 
     def task_id(self) -> str:
--- a/test/unit_tests/test_braket_quantum_task.py
+++ b/test/unit_tests/test_braket_quantum_task.py
@@ -128,13 +128,40 @@ class TestBraketQuantumTask(TestCase):
             tasks=[mock_aws_quantum_task],
             shots=10,
         )
-        mock_aws_quantum_task.return_value.queue_position.return_value = QuantumTaskQueueInfo(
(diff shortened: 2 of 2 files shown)


MohammedAlkindi:
Pushed a test for the coverage failure. It was a consequence of this diff: replacing `AwsQuantumTask(self.task_id()).queue_position()` with `task.queue_position()` removed the last caller of `task_id()`, so line 165 stopped being executed and the 100% floor caught it.

`task_id()` returning what it was constructed with is worth asserting on its own terms, so the test covers both the single-task id and the semicolon-joined batch one.

```
coverage run --source qiskit_braket_provider --parallel-mode -m pytest
coverage combine && coverage report --show-missing --fail-under=100
  596 passed, 4 skipped    TOTAL  2119  0  100%    exit 0
```

Correcting the body: the `numba` DLL block it mentions does not apply in this environment. `numba` 0.67.0 imports and the full suite runs, which produced the numbers above.

